PR #15, opened 2015-06-20
Add new rule handle-done-callback

lo1tuma:
For now the rule performs a really simple check and may not cover each
possible use-case.

* It doesn’t include branch-tracking to make sure that the `done` callback is called in every possible branch.
* It doesn’t check **HOW** the callback is handled, so e.g. `return done;` would be sufficient.

### DIFF
--- a/docs/rules/handle-done-callback.md
+++ b/docs/rules/handle-done-callback.md
@@ -1,0 +1,68 @@
+# Enforces handling of callbacks for async tests (no-exclusive-tests)
+
+Mocha allows you to write asynchronous tests by adding a `done` callback to the parameters of your test function.
+It is easy to forget calling this callback after the asynchronous operation is done.
+
+Example:
+
+```js
+
+it('should work', function (done) {
+    fetchData(options, function (error, data) {
+        expect(error).not.to.be.ok;
+        expect(data).to.deep.equal({ foo: 'bar' });
+    });
+});
+```
+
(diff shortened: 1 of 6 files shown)


docs/rules/handle-done-callback.md
```diff
@@ -0,0 +1,68 @@
...
+It is easy to forget calling this callback after the asynchronous operation is done.
+
+Example:
+
+```js
+
+it('should work', function (done) {
+    fetchData(options, function (error, data) {
+        expect(error).not.to.be.ok;
+        expect(data).to.deep.equal({ foo: 'bar' });
```
lxanders: You could add a comment below this line like `// The done callback was not called` to make this even more clear.

docs/rules/handle-done-callback.md
```diff
@@ -0,0 +1,68 @@
...
+    });
+});
+```
+
+In this example the `done callback was never called and test will time out.
```
lxanders: Close the backtick after `done`

lxanders:
I really like this rule :+1:! It seems to work and everything looks fine (apart from my small comments).

lo1tuma:
Comments addressed :wink:.

I think the two points I mentioned in the PR description could be done in the future (we should file an issue for each). Maybe a generic branch-tracking approach will be implemented directly in ESlint.

PR #23, opened 2015-07-09
Improve detection if done callback is handled

### LINKED ISSUE #20: handle-done-callback: Add checks how the callback is handled specifically (e.g. for simple to detect erroneous calls)
E.g. `return done;` should cause a warning while `return otherImportantFunctionThatHandlesDone(done);` should not.

lo1tuma:
Fixes #20

### DIFF
--- a/lib/rules/handle-done-callback.js
+++ b/lib/rules/handle-done-callback.js
@@ -43,13 +43,23 @@ module.exports = function (context) {
         });
     }
 
+    function isReferenceHandled(reference) {
+        var parent = context.getNodeByRangeIndex(reference.identifier.range[0]).parent;
+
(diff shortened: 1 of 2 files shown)


lib/rules/handle-done-callback.js
```diff
@@ -1,18 +1,43 @@
...
+            },
+            leave: function (node) {
+                if (node === result) {
+                    this.break();
+                }
+            }
+        });
+
+        return result;
+    }
```
lo1tuma: This function is copied and slightly modified from eslint. This can be removed with the next eslint version which will include this changes.

lo1tuma:
This is ready to go.

lxanders:
LGTM :+1:

PR #25, opened 2015-08-11
ESLint 1.x compatibility

2fast2fourier:
ESLint 1.0/1.1 made a few breaking changes, I've updated everything I found.

- Several `.eslintrc` rules changed, mostly name changes or breaking up features. I think I updated these without changing any expected behaviors.
- `ESLintTester` was replaced with `RuleTester`, this did not change the test structure itself.
- Column numbering changed to be 1-indexed, meaning all existing column tests were off-by-one.

Let me know if I missed anything. The tests are passing and it still has 100% coverage.


lo1tuma:
LGTM.

Thank you!

PR #26, opened 2015-08-12
Add no-synchronous-tests rule

2fast2fourier:
I've had a need for a rule based on my experience writing a bunch of async tests within mocha. This rule checks that either a callback was specified or that a promise was returned in the test function.

Mocha dynamically determines whether a test is synchronous by whether the function includes a callback or returns a promise. Because of that, it's easy to end up in a situation where you intend to return a promise (and do not have a `done` callback) but accidentally omit the `return` statement. This leads to a false-positive where the test would automatically succeed regardless of the actual async result.

This rule works well side-by-side with `handle-done-callback`, as that rule will enforce whether or not a callback was actually called.

Let me know if you see any issues with this PR and I'll try to resolve them. Thanks in advance!

### DIFF
--- a/lib/rules/no-synchronous-tests.js
+++ b/lib/rules/no-synchronous-tests.js
@@ -1,0 +1,71 @@
+'use strict';
+
+var _ = require('lodash');
+
+module.exports = function (context) {
+    var possibleAsyncFunctionNames = [
+        'it',
+        'it.only',
+        'test',
+        'test.only',
+        'before',
+        'after',
+        'beforeEach',
+        'afterEach'
+    ];
+
+    function getCalleeName(callee) {
+        if (callee.type === 'MemberExpression') {
+             return callee.object.name + '.' + callee.property.name;
+        }
+
+        return callee.name;
+    }
+
+    function hasParentMochaFunctionCall(functionExpression) {
+        var name;
+
+        if (functionExpression.parent && functionExpression.parent.type === 'CallExpression') {
+            name = getCalleeName(functionExpression.parent.callee);
+            return possibleAsyncFunctionNames.indexOf(name) > -1;
+        }
+
+        return false;
+    }
+
+    function hasAsyncCallback(functionExpression) {
+        return functionExpression.params.length === 1;
+    }
+
+    function findPromiseReturnStatement(nodes) {
+      return _.find(nodes, function (node) {
+        return node.type === 'ReturnStatement' && node.argument && node.argument.type !== 'Literal';
+      });
+    }
+
+    function checkPromiseReturn(functionExpression) {
+        var bodyStatement = functionExpression.body;
+        var returnStatement = null;
+        if (bodyStatement.type === 'BlockStatement') {
+            returnStatement = findPromiseReturnStatement(functionExpression.body.body);
+        } else if (bodyStatement.type === 'CallExpression') {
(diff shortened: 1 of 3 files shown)


lo1tuma:
Thanks for the PR.

Can you explain why a eslint rule would be better instead of using the [`--async-only` option](http://mochajs.org/#usage) from mocha? IMHO the only advantage would be to detect synchronous tests earlier through static analysis while the `--async-only` option would make synchronous tests fail during the test-run. Is there any other advantage?

phated:
The `--async-only` option doesn't work for returning promises in the current stable release.  It has been fixed in master (see https://github.com/mochajs/mocha/pull/1490) but there is no timeline for when that will see a release.  It is also just nice to catch this during the linting process (especially seeing it in your editor)

2fast2fourier:
Sorry, I thought I replied earlier and just realized it didn't go through. Thanks for bringing up `--async-only`, I wasn't aware of it until now. I think this rule still has value as it would allow for someone to catch a mistake before they attempt to run the tests, I find sublime eslint integration to be a big timesaver.

lo1tuma:
Thanks for the explanation, sounds reasonable to me :wink:.

lib/rules/no-synchronous-tests.js
```diff
@@ -0,0 +1,71 @@
...
+        var returnStatement = null;
+        if (bodyStatement.type === 'BlockStatement') {
+            returnStatement = findPromiseReturnStatement(functionExpression.body.body);
+        } else if (bodyStatement.type === 'CallExpression') {
+            //  allow arrow statements calling a promise with implicit return.
+            returnStatement = bodyStatement;
+        }
+
+        if (!returnStatement) {
+            context.report(functionExpression, 'Expected test to handle a callback or return a promise.');
```
lo1tuma: I think the error message could be improved. The rule doesn’t check if a callback is handled, so this is not quite correct. What do you think about `"Unexpected synchronous test."`?

test/rules/no-synchronous-tests.js
```diff
@@ -0,0 +1,69 @@
...
+        {
+            code: 'it("", (done) => { done(); });',
+            ecmaFeatures: { arrowFunctions: true }
+        },
+        {
+            code: 'it("", () => { return promise(); });',
+            ecmaFeatures: { arrowFunctions: true }
+        },
+        {
+            code: 'it("", () => promise() );',
```
lo1tuma: Can you add a similar testcase with a `MemberExpression`?

```js
it("", () => promise.then() );
```

test/rules/no-synchronous-tests.js
```diff
@@ -0,0 +1,69 @@
...
+
+ruleTester.run('no-synchronous-tests', rule, {
+    valid: [
+        'it();',
+        'it("");',
+        'it("", function () { return promise(); });',
+        'it("", function () { return promise(); });',
+        'it("", function () { var promise = myFn(); return promise; });',
+        'var someFn = function(){ }; it("", someFn);',
+        'it("", function (done) { done(); });',
```
lo1tuma: Can you add this testcase as well `it("", function (done) { });`?

2fast2fourier:
I've added those test cases and updated the messaging. Thanks for the feedback!

lo1tuma:
LGTM.

Thanks for your work :beers:.

PR #34, opened 2015-09-13
Implement autofix for no-exclusive-tests

lo1tuma:


lib/rules/no-exclusive-tests.js
```diff
@@ -23,12 +23,22 @@ module.exports = function (context) {
            isPropertyNamedOnly(callee.property);
     }
 
+    function createAutofixFunction(callee) {
+        return function removeOnlyProperty(fixer) {
+            return fixer.removeRange([ callee.object.range[1], callee.range[1] ]);
```
lxanders: This could be improved by extracting variables for the range to make clear, what `callee.object.range[1]` & `callee.range[1]` exactly are but something as general as an `only` function is really hard to name I guess. So unfortunately nothing came to my mind. Do you have an idea?

test/rules/no-exclusive-tests.js
```diff
@@ -26,43 +26,53 @@ ruleTester.run('no-exclusive-tests', rules['no-exclusive-tests'], {
...
         {
             code: 'suite["only"]()',
-            errors: [ { message: expectedErrorMessage, column: 7, line: 1 } ]
+            errors: [ { message: expectedErrorMessage, column: 7, line: 1 } ],
+            output: 'suite()'
         },
         {
             code: 'test.only()',
-            errors: [ { message: expectedErrorMessage, column: 6, line: 1 } ]
+            errors: [ { message: expectedErrorMessage, column: 6, line: 1 } ],
```
lxanders: Could explain to me the counting mechanism for the `column`? For `test.only` and `test["only"]` the number is in both cases `6` here, while these definitely don't have the same starting column of the `o`.

lo1tuma: We are reporting the start position of a `MemberExpression` property. For `test["only"]` we have a computed `MemberExpression` with the string literal `"only"` as the property. For `test.only` we have a non-computed `MemberExpression` with the identifier `only` as the property.

You can play around with the [esprima online demo](http://esprima.org/demo/parse.html) (don’t forget to enable the `Line and column-based` option).

lib/rules/no-exclusive-tests.js
lo1tuma: done

test/rules/no-exclusive-tests.js
lxanders: Thanks for the clarification and the hint to the demo :+1:

lib/rules/no-exclusive-tests.js
```diff
@@ -23,12 +23,26 @@ module.exports = function (context) {
            isPropertyNamedOnly(callee.property);
     }
 
+    function createAutofixFunction(callee) {
+        var endRangeOfMemberExpression = callee.range[1],
+            endRangeOfMemberExpressionObject = callee.object.range[1],
+            rangeToRemove = [ endRangeOfMemberExpressionObject, endRangeOfMemberExpression ];
```
lxanders: :+1:

lxanders:
Looks good for me. I will wait for the coveralls check (which is not yet done) and try the feature itself tomorrow and then merge :beer:

lxanders:
Did a short check and it seems to work (tested it with a `describe.only`). Merging :+1:

Issue #48, opened 2016-01-18
Rule proposal : Disallow arrow functions as parameters of mocha functions


rik:
It's discouraged: http://mochajs.org/#arrow-functions

lo1tuma:
Sounds reasonable to me.

lo1tuma:
Implemented in #78 and released with version `4.1.0`.

PR #49, opened 2016-02-11
Update to eslint 2.0.0

willyaranda:
Hi there!

I am working with latest eslint and wanted to use this as a plugin. This is a PR to fix it for the upcoming 2.0 of ESlint.

It should have a major bump of version number. But that can be done once #48 is done (which I can happily try to do).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,7 +4,10 @@
     },
 
     "parserOptions": {
-        "ecmaVersion": 6,
+      "ecmaVersion": 5,
+      "ecmaFeatures": {
+        "globalReturn": true
+      }
     },
 
     "rules": {
@@ -23,6 +26,8 @@
         "no-dupe-keys": 2,
         "no-dupe-args": 2,
         "no-else-return": 2,
+        "no-empty": 2,
+        "no-empty-character-class": 2,
         "no-eq-null": 2,
         "no-eval": 2,
         "no-ex-assign": 2,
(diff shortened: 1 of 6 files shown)


.eslintrc
```diff
@@ -3,8 +3,8 @@
         "node": true
     },
 
-    "ecmaFeatures": {
-        "globalReturn": false
+    "parserOptions": {
+        "ecmaVersion": 6,
```
lo1tuma: Currently this project still uses ES5. Also I really want to keep `globalReturn` disabled.

.eslintrc
```diff
@@ -23,9 +23,6 @@
         "no-dupe-keys": 2,
         "no-dupe-args": 2,
         "no-else-return": 2,
-        "no-empty": 2,
-        "no-empty-character-class": 2,
```
lo1tuma: `no-empty` and `no-empty-character-class` still exist is eslint 2. What is the reason to remove the config of those rules here?

.eslintrc
```diff
@@ -23,9 +23,6 @@
...
-        "no-empty-label": 2,
```
lo1tuma: According to the changelog this rule has been replaced by `no-labels`  with {"allowLoop": true, "allowSwitch": true} option.

lo1tuma:
Thanks for the Pull Request.

Most of the changes looking good so far, I’ve only spotted a few minor problems.
I don’t plan to release a new major version based on a release candidate. I could only offer to release a beta or release candidate version until ESLint 2.0 is out.

willyaranda:
ESlint 2 has just been released. I updated the PR with the comments you made. I hope you find this useful.

lo1tuma:
Awesome. Thank you!

:beers:

PR #50, opened 2016-02-23
switch to lodash/fp

phated:
lodash now ships with it's functional counterpart, how would you feel about switching back?

Also, lodash ships with it's modular build so I could switch this to `var find = require('lodash/fp/find')` if you think that looks better.

### DIFF
--- a/lib/rules/handle-done-callback.js
+++ b/lib/rules/handle-done-callback.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var R = require('ramda');
+var L = require('lodash/fp');
 
 module.exports = function (context) {
     var possibleAsyncFunctionNames = [
@@ -38,7 +38,7 @@ module.exports = function (context) {
     }
 
     function findParamInScope(paramName, scope) {
-        return R.find(function (variable) {
+        return L.find(function (variable) {
             return variable.name === paramName && variable.defs[0].type === 'Parameter';
         }, scope.variables);
     }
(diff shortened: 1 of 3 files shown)


lxanders:
Thanks for your input!

lodash seems to have come far into the direction of ramda (offering a similar amount of e.g. currying niceness). ramda has this too however. IMHO it comes down to preferences. Personally I don't see a need to change right now. @lo1tuma might think different about this and I would not stand in the way - both libraries are awesome :+1:.

jdalton:
@lxanders It's current usage in eslint-plugin-mocha isn't taking advantage of currying or any of the things that makes either FP lib shine. As is it's aligned closer with traditional lodash than the other flavors.

lib/rules/handle-done-callback.js
```diff
@@ -38,7 +38,7 @@ module.exports = function (context) {
     }
 
     function findParamInScope(paramName, scope) {
-        return R.find(function (variable) {
+        return L.find(function (variable) {
             return variable.name === paramName && variable.defs[0].type === 'Parameter';
         }, scope.variables);
     }
```
jdalton:
```js
L.find({ name: paramName, defs: [{ type: 'Parameter' }] }, scope.variables);
```

gilligan:
@jdalton @phated you guys really are desperate.

jdalton:
@gilligan Naw.

lo1tuma:
Thanks for the pull request. 

I don't really see any value in switching to lodash. My personal preference is ramda which I also use a lot in some of my other projects.

jdalton:
Cool. cool.

phated:
Thanks @lo1tuma - I made the fork to use in my own projects since lodash dedupes much better than ramda and I figured I'd send it as a PR, just in case.

PR #55, opened 2016-04-08
Add rule no-skipped-tests

treythomas123:
Add a rule to catch skipped tests (e.g. `describe.skip()` or `xdescribe`).

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -1,6 +1,7 @@
 # Rules
 
 * [no-exclusive-tests](no-exclusive-tests.md) - disallow exclusive mocha tests (fixable)
+* [no-skipped-tests](no-skipped-tests.md) - disallow skipped mocha tests (fixable)
 * [handle-done-callback](handle-done-callback.md) - enforces handling of callbacks for async tests
 * [no-synchronous-tests](no-synchronous-tests.md) - disallow synchronous tests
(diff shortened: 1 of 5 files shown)


lo1tuma:
I wonder if we should name the rule `no-pending-tests` instead? According to [the mocha documentation](http://mochajs.org/#inclusive-tests) skipped/inclusive tests are reported as pending:

> Anything skipped will be marked as pending, and reported as such.

What do you think?

lo1tuma:
Apart from that, the code looks good to me.

treythomas123:
I think you'd want to handle them differently, because I think an unimplemented (pending) test has a different meaning than an implemented, but skipped test. The mocha docs say:

> "Pending"--as in "someone should write these test cases eventually"--test-cases are simply those without a callback

Which is different from "skipped", where there's an implementation but you don't want to run it for whatever reason. There's actually [an open issue](https://github.com/mochajs/mocha/issues/1815) in mocha for reporting them separately instead of lumping skipped tests under "pending", which I agree with.

lo1tuma:
That makes totally sense to me. I hope the mocha issue will be fixed for the next major release.

Thanks for your contribution and the good discussion... merging 🍻 .

treythomas123:
Cool, thanks!

Issue #51, opened 2016-02-24
Rule proposal: no pending tests


catkins:
Inside a large mocha test suite, it's quite difficult to hunt down pending specs of PRs past. It would be awesome to be able to mark them as warnings to pester each other to implement or delete them.

#### OK

```jsx
describe('some module', function() {
  it('implements a test', function() {
    expect(true).to.be.ok;
  });
});
```

### Error

```jsx
describe('some module', function() {
  it('does not yet implement a test');

  it.skip('implemented but skipped', function() {
    expect(false).to.be.ok;
  });
});

describe.skip('another module', function() {
  // ...
});
```

lo1tuma:
Sounds good to me.

lo1tuma:
Implemented in #59 and released with version 2.2.0.

PR #59, opened 2016-04-13
Add rule no-pending-tests

treythomas123:
Implements [#51](https://github.com/lo1tuma/eslint-plugin-mocha/issues/51)

### DIFF
--- a/docs/rules/no-pending-tests.md
+++ b/docs/rules/no-pending-tests.md
@@ -1,0 +1,43 @@
+# Disallow Pending Tests (no-pending-tests)
+
+Mocha allows specification of pending tests, which represent tests that aren't yet implemented, but are intended to be implemented eventually. These are designated like a normal mocha test or suite, but with only the first argument provided (no callback for the actual implementation). For example: `describe('unimplemented bdd suite');`
+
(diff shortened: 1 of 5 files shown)


docs/rules/no-pending-tests.md
```diff
@@ -0,0 +1,41 @@
+# Disallow Pending Tests (no-pending-tests)
+
+Mocha allows specification of pending tests, which represent tests that aren't yet implemented, but are intended to be implemented eventually. These are designated like a normal mocha test or suite, but with only the first argument provided (no callback for the actual implementation). For example: `describe('unimplemented bdd suite');`
```
lo1tuma: AFAIK mocha currently doesn’t support pending suites. If the second argument of `describe` is missing an error will be thrown.
There is an [upstream issue](https://github.com/mochajs/mocha/issues/1744) about this issue, but as I understand it, it hasn’t been decided if pending suites will be implemented or if the error message will be improved.

treythomas123: Oh I see -- just added a new commit to remove the handling for those. I guess I just assumed mocha supported pending suites.

lo1tuma:
Awesome, thank you 👍

PR #63, opened 2016-06-01
Remove autofix on no-exclusive-tests rule. 

### LINKED ISSUE #60: Don't autofix no-exclusive-tests
With autofix eslint is becoming a great tool to fix stylistic linting issues in code. I love having `--fix` on save in Atom's `linter-eslint`.

While I love `no-exclusive-tests` to make sure I don't commit `.only` (pre-commit hook). Having autofix removing `.only` on save is obviously undesired.

Right now the only solution for this is disabling the `no-exclusive-tests` rule. It would be great to have autofix be opt-in for this rule.

wietsevenema:
Thanks for this project! 

I would like to suggest an improvement that is relevant to me: I have setup my editor to auto-fix eslint errors on save. When the exclusive tests rule is auto-fixable, I lose the ability to single out tests. 

Fixes #60

### DIFF
--- a/lib/rules/no-exclusive-tests.js
+++ b/lib/rules/no-exclusive-tests.js
@@ -24,16 +24,6 @@ module.exports = function (context) {
            isPropertyNamedOnly(callee.property);
     }
 
-    function createAutofixFunction(callee) {
-        var endRangeOfMemberExpression = callee.range[1],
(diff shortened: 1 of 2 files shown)


lo1tuma:
LGTM, thanks @wietsevenema.

PR #68, opened 2016-06-29
feat(rules): add 'valid-test-description' rule

### LINKED ISSUE #66: Rule proposal: enforce it() description to start with "should"
Would it make sense to enforce having descriptive test names by warning if `it()` description does not start with "should". Example of violation:

```
it("I want to test something", function () {
  // ...
});
```

Valid usage:

```
it("should test something", function () {
  // ...
});
```

Thanks.

alecxe:
This should close #66. Will create a separate similar issue for describe/suite.

Please review and feel free to criticize/improve/discuss. Thanks!


docs/rules/valid-it-test-desc.md
```diff
@@ -0,0 +1,31 @@
+# Check "it()" and "test()" descriptions against a pre-configured regular expression
+
+This rule enforces the "it()" and "test()" descriptions to follow the desired format.
```
lo1tuma: `specify` is another alias for `it`.
I’m not saying that we need to implement support for this alias now, but maybe we should rethink the name of the rule to make future additions possible without changing the rule name.
Maybe we can call it just `valid-description` then we could easily add support for e.g. `describe` in the future. What do you think?

alecxe: @lo1tuma good idea, let's make a more generic rule title and a configurable list of aliases. How about something like `valid-test-description` and, for a similar rule for `describe` - `valid-suite-description`?.. Thanks.

lo1tuma: Sounds good to me 👍 .

alecxe:
@lo1tuma please re-review. Thanks.

lo1tuma:
LGTM.

Thanks for your contribution 🍻 .

PR #74, opened 2016-07-03
feat(rules): add 'valid-suite-description' rule

### LINKED ISSUE #69: Rule proposal: add "valid-suite-description" rule
Similar to the existing [valid-test-description](https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/lib/rules/valid-test-description.js) rule, but to match against the suite descriptions (`describe` or `context` or `suite`). Thanks.

alecxe:
Should close #69. Thanks.

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -7,3 +7,4 @@
 * [no-synchronous-tests](no-synchronous-tests.md) - disallow synchronous tests
 * [no-global-tests](no-global-tests.md) - disallow global tests
 * [valid-test-description](valid-test-description.md) - match test descriptions against a pre-configured regular expression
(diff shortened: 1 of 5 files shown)


alecxe:
Btw, `valid-test-description` and `valid-suite-description` rules are quite similar in their implementation. To avoid code duplication we should probably refactor these rules and extract similar code chunks into reusable exported functions. I would though address after both rules are present in the codebase. What do you think? Thanks.

lo1tuma:
Code LGTM. It seems like there are some merge conflicts. Can you rebase and resolve those conflicts?

Extracting duplicated code into re-useable libs would be very nice. I was also planning to do this for some other rules in this plugin but haven’t found time yet.

alecxe:
Okay, fixed the conflicts. Thanks.

lo1tuma:
Thank you.

PR #78, opened 2016-07-12
no-mocha-global-arrows: New rule

paulmelnikow:
For #48

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -8,3 +8,4 @@
 * [no-global-tests](no-global-tests.md) - disallow global tests
 * [valid-test-description](valid-test-description.md) - match test descriptions against a pre-configured regular expression
 * [valid-suite-description](valid-suite-description.md) - match suite descriptions against a pre-configured regular expression
+* [no-mocha-arrows](no-mocha-arrows.md) - disallow arrow functions as arguments to mocha globals
--- a/docs/rules/no-mocha-arrows.md
+++ b/docs/rules/no-mocha-arrows.md
@@ -1,0 +1,31 @@
+# Disallow Arrow Functions as Arguments to Mocha Functions (no-mocha-arrows)
+
+Mocha [discourages](http://mochajs.org/#arrow-functions) passing it arrow functions as arguments. This rule prevents their use on the Mocha globals.
+
+## Rule Details
+
+This rule looks for occurrences of the Mocha functions (`describe`, `it`, `beforeEach`, etc.) within the source code.
+
+The following patterns are considered warnings:
+
+```js
+it(() => { assert(something, false); })
+it("should be false", () => { assert(something, false); })
+beforeEach(() => { doSomething(); })
+beforeEach((done) => { doSomething(); done(); })
+```
+
+These patterns would not be considered warnings:
+
+```js
+it()
+it(function() { assert(something, false); })
+it("should be false", function() { assert(something, false); })
+```
+
+This does not check usage of the [`require` interface](http://mochajs.org/#require) for Mocha, only the globals.
+
+## When Not To Use It
+
+* If you want to pass arrow functions to mocha, turn this rule off.
+* If you have other globals which share names with mocha globals, you should turn this rule off, because it would raise warnings.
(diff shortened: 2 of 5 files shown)


lib/rules/no-mocha-global-arrows.js
```diff
@@ -0,0 +1,51 @@
...
+    function isLikelyMochaGlobal(scope, name) {
+        return !R.find(R.propEq('name', name), scope.variables);
+    }
+
+    return {
+        CallExpression: function (node) {
+            var callee = node.callee,
+                fnArg;
+
+            if (callee && callee.name && mochaFunctionNames.indexOf(callee.name) > -1) {
```
lo1tuma: This doesn't work for function names like `it.only`. See how this is done in [other rules](https://github.com/lo1tuma/eslint-plugin-mocha/blob/2f560171dd691ae444eb0a7ea819374a9497df50/lib/rules/handle-done-callback.js#L19-L25).

lib/rules/no-mocha-global-arrows.js
```diff
@@ -0,0 +1,51 @@
...
+                fnArg = node.arguments.slice(-1)[0];
```
lo1tuma: Wouldn’t it be easier to just use `node.arguments[1]`?

test/rules/no-mocha-global-arrows.js
```diff
@@ -0,0 +1,45 @@
...
+    expectedErrorMessage = 'Do not pass arrow functions to it()';
+
+ruleTester.run('no-mocha-global-arrows', rules['no-mocha-global-arrows'], {
+
+    valid: [
+        'it()',
+        'it(function() { assert(something, false); })',
+        'it("should be false", function() { assert(something, false); })',
+        {
+            // In this example, `it` is not a global.
```
lo1tuma: Mocha also has an [`require`](http://mochajs.org/#require) interface in which you would require mocha functions. In this case they would not be globals so I wonder if we should support this interface too?
What do you think?

lib/rules/no-mocha-global-arrows.js
paulmelnikow: Fixed, thanks: b18e8ba

lib/rules/no-mocha-global-arrows.js
paulmelnikow: No, that won't work. Consider these examples:

```js
beforeEach(function () { });
beforeEach('prepare', function () { });
```

In the first example, `node.arguments[1]` will be undefined.

test/rules/no-mocha-global-arrows.js
paulmelnikow: I don't have a need for it, but it seems like a nice idea.

Doing it well would require more complex logic. First, you'd need to look for `require('mocha')`. Then you'd need to see `var darkBrown = require('mocha')` and then check `darkBrown.anything`. And see `var b4 = require('mocha').beforeEach` and check `b4`.

If you think we should support it in the future, I'd be happy to adjust the interface. Instead of `no-mocha-global-arrows`, it could be `no-mocha-arrows`. Initially it would just check globals, as this does, but if someone wanted to implement the require syntax, they could add a `'globals': true` option which preserves this behavior, and a `'globals': false, 'require': true` option to turn on the new logic. (`'require': true` might not be needed. The only reason I would include it is if it slows down the checking.)

It would share a few lines of code and `mochaFunctionNames` with this rule; otherwise I'd suggest using a separate rule.

lib/rules/no-mocha-global-arrows.js
lo1tuma: Got it, that makes sense.

test/rules/no-mocha-global-arrows.js
lo1tuma: Personally I also don’t have a need for this. So let’s rename the rule as you suggested, then we are safe for future extension of this rule 👍.

paulmelnikow:
Okay, renamed and rebased.

paulmelnikow:
@lo1tuma Can you merge this?

lo1tuma:
@paulmelnikow Sorry for the late response, I was traveling.

Thanks for your contribution, LGTM.

paulmelnikow:
Thanks!

PR #79, opened 2016-07-15
Add support for ember-mocha methods

job13er:
The `ember-mocha` project adds a few new `describe()` methods:
 * `describeComponent()`
 * `describeModel()`
 * `describeModule()`

This change makes the lint rules aware of these functions as well.

### DIFF
--- a/lib/rules/no-exclusive-tests.js
+++ b/lib/rules/no-exclusive-tests.js
@@ -4,6 +4,9 @@ module.exports = function (context) {
     var mochaTestFunctions = [
         'it',
         'describe',
+        'describeComponent',
+        'describeModel',
+        'describeModule',
         'suite',
         'test',
         'context',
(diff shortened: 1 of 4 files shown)


alexlafroscia:
This getting accepted would relieve my need for #80

lo1tuma:
I don’t like the idea of supporting specific extensions of mocha in this plugin; maintenance would become much harder.
So I would say we should go with the more generic approach as suggested in #80. When we use a [shared setting](http://eslint.org/docs/user-guide/configuring#adding-shared-settings) we could even get support for additional mocha methods in all rules of this plugin.

alexlafroscia:
If #81 is accepted, this would be done by:

```json
{
    "rules": {
        "mocha/no-exclusive-tests": ["error", {
            "additionalTestFunctions": [
                "describeComponent",
                "describeModel"
                "describeModule"
            ]
        }]
    }
}
```

lo1tuma:
Closing in favor of #81.

PR #81, opened 2016-07-24
Allow custom test functions

### LINKED ISSUE #80: Feature Request: allow users to provide variables to protect from `.only`
I'm using `ember-cli-mocha` which provides some functions that build on top of `describe`, such as `describeModule`.  I would love to be able to provide an array of additional things to disallow `.only` from being called.

alexlafroscia:
This feature is useful if using this plugin with a package like `ember-mocha`, which provides its own set of `describe` functions. A custom array can be provided so that those extensions can also be protected from having `only` called.

One thought about my changes is that it introduces a configuration object rather than an array.  We could make the option parameter just an array, but I kind of like the extra context provided by the named keys.  The difference is basically between this (how I have it right now)

```json
{
     "rules": {
         "mocha/no-exclusive-tests": ["error", {
             "additionalTestFunctions": [
                 "describeModule"
             ]
         }]
     }
 }
```

and this (which I think provides less context and it more confusing)

```json
{
     "rules": {
         "mocha/no-exclusive-tests": ["error", [
             "describeModule"
         ]]
     }
 }
```

Closes #80

### DIFF
--- a/docs/rules/no-exclusive-tests.md
+++ b/docs/rules/no-exclusive-tests.md
@@ -51,18 +51,19 @@ test.skip("bar", function () {});
 
 # Options
 
-This rule supports the following configuration options, which can be passed in as a single object.  An example is below.
+This rule supports the following shared configuration options:
 
 * `additionalTestFunctions`: An array of extra test functions to protect.  This might be used with a custom Mocha extension, such as [`ember-mocha`](https://github.com/switchfly/ember-mocha)
 
 ```json
 {
     "rules": {
-        "mocha/no-exclusive-tests": ["error", {
-            "additionalTestFunctions": [
-                "describeModule"
-            ]
-        }]
+        "mocha/no-exclusive-tests": "error"
+    },
+    "settings": {
+       "mocha/additionalTestFunctions": [
+           "describeModule"
+       ]
     }
 }
 ```
--- a/docs/rules/no-exclusive-tests.md
+++ b/docs/rules/no-exclusive-tests.md
@@ -49,6 +49,24 @@ suite.skip("bar", function () {});
 test.skip("bar", function () {});
 ```
 
+# Options
+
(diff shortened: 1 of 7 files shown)


alexlafroscia:
Sorry about the messy Coveralls comments, I overwrote my commits a few times.

lo1tuma:
Thanks for the Pull Request.

It seems like adding those options to each rule introduces a lot of duplication, especially if we add more rules to this plugin in the future.
What do you think about using shared settings (e.g. as [`eslint-plugin-import`](https://github.com/benmosher/eslint-plugin-import) or [`eslint-plugin-backbone`](https://github.com/ilyavolodin/eslint-plugin-backbone) does)?
With shared settings the configuration would look like this:

```json
{
     "rules": {
         "mocha/no-exclusive-tests": "error",
         "mocha/no-skipped-tests": "error"
     },
    "settings": {
        "mocha/additionalTestFunctions": [ "describeModule" ],
        "mocha/additionalXFunctions": [ "xcustom" ]
    }
 }
```

or

```json
{
     "rules": {
         "mocha/no-exclusive-tests": "error",
         "mocha/no-skipped-tests": "error"
     },
    "settings": {
        "mocha": {
            "additionalTestFunctions": [ "describeModule" ],
            "additionalXFunctions": [ "xcustom" ]
        }
    }
 }
```

alexlafroscia:
Huh, that's really cool.  I didn't know you could configure ESLint that way -- I like it much better.  Totally agree that the current approach means duplication, I'll fix up this PR to implement with shared settings instead.

alexlafroscia:
Added the shared config just like @lo1tuma suggested, allowing either the one-line or nested option object.

lo1tuma:
Great work @alexlafroscia.

Thank you 🍻 .

PR #85, opened 2016-07-26
Add rule `no-sibling-hooks` (fixes #82)

### LINKED ISSUE #82: Rule proposal: No multiple hooks on the same level
In the codebase of one of my projects, I often see multiple hooks of the same type being defined under the same `describe` level:

```js
describe('foo', function() {
  before(function() {
    // set up something
  });
  before(function() {
    // set up something else
  });

  // tests
});
```

This can be confusing when one hook is before the tests and the other after the tests (think `after`/`afterEach`). Anyway, this is IMO not a good pattern as you can and should do everything in one hook.

jfmengels:
Add rule `no-sibling-hooks` (fixes #82)

Let me know if you want me to change something or if I've missed something :)

I deactivatd two rules [here](https://github.com/lo1tuma/eslint-plugin-mocha/pull/85/files#diff-23841b5a313e8e46a51f3d758eaa6f4fR22) about the function complexity. I could not make the function less complex. I could move the last part of the function to another function, but I'm not sure it would make it less complex, as you'd pass a few arguments around. Not opposed to doing that though. Let me know what you wish for me to do.

### DIFF
--- a/docs/rules/no-sibling-hooks.md
+++ b/docs/rules/no-sibling-hooks.md
@@ -1,0 +1,59 @@
+# Disallow hooks (no-hooks)
+
+Mocha proposes hooks that allow code to be run before or after every or all tests. This helps define a common setup or teardown process for every test.
+It is possible to declare a hook multiple times inside the same test suite, but it can confusing. It is better to have one hook handle the whole of the setup or teardown logic of the test suite.
+
+## Rule Details
+
+This rule looks for every call to `before`, `after`, `beforeEach` and `afterEach` and reports them if they are called at least twice inside of a test suite at the same level.
+
+The following patterns are considered warnings:
+
+```js
+describe('foo', function () {
+    var mockUser;
+    var mockLocation;
+
+    before(function () { // Is allowed this time
+        mockUser = {age: 50};
+    });
+
+    before(function () { // Duplicate! Is not allowed this time
+        mockLocation = {city: 'New York'};
+    });
+
+    // Same for the other hooks
+    after(function () {});
+    after(function () {}); // Duplicate!
+
+    beforeEach(function () {});
+    beforeEach(function () {}); // Duplicate!
+
+    afterEach(function () {});
+    afterEach(function () {}); // Duplicate!
+});
+```
+
+These patterns would not be considered warnings:
+
+```js
+describe('foo', function () {
+    var mockUser;
+    var mockLocation;
+
+    before(function () { // Is allowed this time
+        mockUser = {age: 50};
+        mockLocation = {city: 'New York'};
+    });
+
+    describe('bar', function () {
+      before(function () { // Is allowed because it's nested in a new describe
+        // ...
+      });
+    });
+});
+```
+
+## When Not To Use It
+
+* If you use another library which exposes a similar API as mocha (e.g. `before`, `after`), you should turn this rule off, because it would raise warnings.
--- a/docs/rules/no-sibling-hooks.md
+++ b/docs/rules/no-sibling-hooks.md
@@ -1,7 +1,7 @@
 # Disallow hooks (no-hooks)
 
 Mocha proposes hooks that allow code to be run before or after every or all tests. This helps define a common setup or teardown process for every test.
-It is possible to declare a hook multiple times inside the same test suite, but it can confusing. It is better to have one hook handle the whole of the setup or teardown logic of the test suite.
+It is possible to declare a hook multiple times inside the same test suite, but it can be confusing. It is better to have one hook handle the whole of the setup or teardown logic of the test suite.
 
 ## Rule Details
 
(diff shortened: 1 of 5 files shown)


docs/rules/no-sibling-hooks.md
```diff
@@ -0,0 +1,59 @@
+# Disallow hooks (no-hooks)
+
+Mocha proposes hooks that allow code to be run before or after every or all tests. This helps define a common setup or teardown process for every test.
+It is possible to declare a hook multiple times inside the same test suite, but it can confusing. It is better to have one hook handle the whole of the setup or teardown logic of the test suite.
```
lo1tuma:
```diff
-but it can confusing
+but it can be confusing
```

lib/rules/no-sibling-hooks.js
```diff
@@ -0,0 +1,54 @@
...
+        },
+
+        CallExpression: function (node) { // eslint-disable-line complexity, max-statements
+            var name;
+            if (node.callee.type !== 'Identifier') {
+              return;
+            }
+
+            name = node.callee.name;
+            if (name === 'describe') {
```
lo1tuma: Should we also support mocha’s TDD interface which uses `suite` instead of `describe` and maybe also the `describe` alias `context`?

Additionally to that should we also support things like `xdescribe`, `xcontext`, `describe.skip`, `describe.only` etc?

test/rules/no-sibling-hooks.js
```diff
@@ -0,0 +1,80 @@
...
+        {
+            code: [
+              'describe(function() {',
+              '    before(function() {});',
+              '    describe(function() {',
+              '        before(function() {});',
+              '        before(function() {});',
+              '    });',
+              '});'
+            ].join(' '),
```
lo1tuma: Maybe it would be easier to join with `\n` instead of a space. Then you wouldn’t end up with an error in column `115`.

💅

jfmengels: True. My reasoning was that it polluted the mocha output by printing one test onto several lines
image

But you're right, as it makes using the correct location harder, so I'll change that.

lib/rules/no-sibling-hooks.js
jfmengels: > Should we also support mocha’s TDD interface which uses suite instead of describe?

I don't know how worth it it is as I've never seen `suite` used.
Actually, I'm noticing there are [multiple interfaces](https://mochajs.org/#interfaces) (making me think there should actually be a rule to enforce a consistent interface among the same project).
Do we want to only handle `suite`, or also the other ones described in that section?

Anyway, if we want to support `suite`, we should handle suite(), test(), suiteSetup(), suiteTeardown(), setup(), and teardown() too as long with `before` and co.

I actually don't know what happens when you use both a `before` and a `suiteSetup()` (or is the equivalent `setup()` ?) Does it break or is that valid and should we then consider them to be aliases and report when both are used?

Or maybe we should keep it this way until someone requests it...? :confused:
If the rest of the plugin handles `suite` correctly, then we should, otherwise, it's pretty pointless unless someone wants to invest time to update every concerned rule.

> and maybe also the describe alias context?

Sure. Again, never heard of this one before. There should probably be a rule enforce the use of `before`/`it` or `context`/`specify`. Unless it's so unknown that no one ever uses it.

> Additionally to that should we also support things like xdescribe, xcontext, describe.skip, describe.only etc?

The double use of `before` inside a `describe.only` is still a valid concern, and I think it would then be easy to generalize it to `xdescribe` and friends.

jfmengels:
> Additionally to that should we also support things like xdescribe, xcontext, describe.skip, describe.only?

Added support for those

> Should we also support mocha’s TDD interface which uses suite instead of describe and maybe also the describe alias context?

Still TBD. I'd vote for no unless someone really wants it or it's properly handled with the other rules. If this is the only rule that will support it, I don't see the point.

jfmengels:
Actually, if we want to do that, we'd probably better update `no-hooks` and similar (existing or upcoming rules), but I'm a bit worried that `setup` for instance is a too common function name, and disallowing that one in `no-hooks` would be an unwanted effect. It would be fine if we can know for certain whether the linted file is a test file or not, but I wouldn't know how to do that.

Or actually, we could probably do it by not reporting the variables that were declared in the file (using `context.getScope()`. Unless setting `mocha: true` in the ESLint env settings makes them available in the scope. Will try to play with this to know if we can have a nice solution.

lo1tuma:
I would be ok with not supporting the TDD interface now. We can add support for it if somebody requests that.

I think the right approach would be check where `setup` and friends were declared (might be also imported with custom name) or if it is a global. But as I said, we can skip this for the moment and wait until somebody reports such a use-case.

jfmengels:
I'm fine with that. Then this PR is up for review

lo1tuma:
LGTM

jfmengels:
cheering-cheer-pom-crab-d3MM4vAejYomlzGg

PR #87, opened 2016-08-01
Add `no-top-level-hooks` rule (fixes #37)

### LINKED ISSUE #37: Rule proposal: forbid top level before / after hooks
Those should be only used in a describe block

jfmengels:
Add `no-top-level-hooks` rule (fixes #37)

Since this rule has a lot of common code with `no-hooks` and `no-sibling-hooks`, I refactored them all to use common methods. It's all much cleaner now, but it might be a lot to look at for one PR. Let me know if you wish for me to split this PR up into two (one for the work on this rule and one for the refactoring)


lo1tuma:
@jfmengels Sorry I totally forgot about this PR. The refactoring looks awesome.

Thank you.

jfmengels:
No problem :)

Btw, in case you want to release this, I still have two PRs incoming (was waiting for these to be merged): no-hooks-for-single-case (#44) and no-return-and-callback (#88, which still needs to be discussed/bikeshedded/approved, but it's been useful to me already)

PR #94, opened 2016-08-08
Add rule `no-return-and-callback` (fixes #88)

### LINKED ISSUE #88: Rule proposal: Forbid tests with both a callback and a return statement
In Mocha v3, it is not allowed to return a Promise in tests while the test function takes a `done` callback function.

It would be nice to detect this, so that you have a tighter feedback loop when coding, instead of waiting for your test to run. It could also help with the migration to v3 (I did this yesterday in a relatively big project, and it was a bit time-consuming).

I'm not sure whether we should forbid (top-level) `return` statements altogether, or only when we can detect that a Promise is returned (with a `.then` call). I think the former would be fine, as there is little value in returning something in a test when you have a callback, and it would cover a lot more cases where we can't know for sure that a Promise is being returned.

### Incorrect

```js
it('should XYZ', function(done) {
  return foo()
    .then(bar)
    .nodeify(done);
});
```

### Correct

```js
it('should XYZ', function(done) {
  foo()
    .then(bar)
    .nodeify(done);
});

it('should XYZ', function() {
  return foo()
    .then(bar);
});
```

As for the rule name, I'm not too sure yet. `no-return-and-callback`? Or if we want to make sure the returned statement is a Promise, `no-promise-and-callback`?

jfmengels:
Add rule `no-return-and-callback` (fixes #88)

Still needs to be validated, but thought I might as well make the PR as I have already written the rule. As said in the issue, I think it's better to warn when we return something, and not only if we can be sure it's a Promise that is being returned.

By the way, this rule will also warn for the same pattern in hooks. The [Mocha changelog](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#boom-breaking-changes) only mentionned this new behavior for tests, but after trying it out, I confirm it also applies to hooks.

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -6,6 +6,7 @@
 * [handle-done-callback](handle-done-callback.md) - enforces handling of callbacks for async tests
 * [no-synchronous-tests](no-synchronous-tests.md) - disallow synchronous tests
 * [no-global-tests](no-global-tests.md) - disallow global tests
+* [no-return-and-callback](no-return-and-callback.md) - disallow returning in a test or hook function that uses a callback
 * [valid-test-description](valid-test-description.md) - match test descriptions against a pre-configured regular expression
 * [valid-suite-description](valid-suite-description.md) - match suite descriptions against a pre-configured regular expression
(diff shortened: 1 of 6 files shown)


lo1tuma:
LGTM.

Thank you.

jfmengels:
:tada:

PR #95, opened 2016-08-08
Add `no-hooks-for-single-case` rule (fixes #44)

### LINKED ISSUE #44: Rule proposal: disallow before, beforeEach, after, afterEach if you have only one test

jfmengels:
Add `no-hooks-for-single-case` rule (fixes #44)

### DIFF
--- a/docs/rules/no-hooks-for-single-case.md
+++ b/docs/rules/no-hooks-for-single-case.md
@@ -1,0 +1,64 @@
+# Disallow hooks for a single test or test suite (no-hooks-for-single-case)
+
+Mocha proposes hooks that allow code to be run before or after every or all tests. This helps define a common setup or teardown process for every test. These hooks are not useful when there is only one test case, as it would then make more sense to move the hooks' operations in the test directly.
+
+## Rule Details
+
+This rule looks for every call to `before`, `after`, `beforeEach` and `afterEach` and reports them if they are called when there is less than two tests and/or tests suites in the same test suite.
+
+The following patterns are considered warnings:
+
+```js
+describe('foo', function () {
+    before(function () { /* ... */ }); // Not allowed as there is only a test suite next to this hook.
+
+    describe('bar', function() {
+      /* ... */
+    });
+});
+
+describe('foo', function () {
+    after(function () { /* ... */ }); // Not allowed as there is only a test case next to this hook.
+
+    it('should XYZ', function() {
+      /* ... */
+    });
+});
+
+describe('foo', function () {
+    beforeEach(function () { /* ... */ }); // Not allowed as there is no test suites or cases next to this hook.
+});
+```
+
+These patterns would not be considered warnings:
+
+```js
+describe('foo', function () {
+    before(function () { /* ... */ });
+
+    it('should XYZ', function() {
+      /* ... */
+    });
+
+    it('should ABC', function() {
+      /* ... */
+    });
+});
+
+describe('foo', function () {
+    before(function () { /* ... */ });
+
+    it('should XYZ', function() {
+      /* ... */
+    });
+
+    describe('bar', function() {
+      /* ... */
+    });
+});
+```
+
+## When Not To Use It
+
+* If you use another library which exposes a similar API as mocha (e.g. `before`, `after`), you should turn this rule off, because it would raise warnings.
+* If you turned `no-hooks` on, you should turn this rule off, because it would raise several warnings for the same root cause.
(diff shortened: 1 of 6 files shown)


lo1tuma:
It turns out that there is one good use-case for using the `after` hook even if there is only one test (which I wasn’t aware of when I created #44).

If a test requires some clean-up code to run even if the test fails then this must be done in the `after` hook because the test function exits on the first failing assertion.

lib/rules/no-hooks-for-single-case.js
```diff
@@ -0,0 +1,56 @@
...
+
+    function popLayer(node) {
+        var layer = layers[layers.length - 1];
+        if (layer.describeNode === node) {
+            if (layer.testCount <= 1) {
+                layer.hookNodes.forEach(function (hookNode) {
+                    context.report({
+                        node: hookNode,
+                        // message: 'Unexpected use of duplicate Mocha `' + name + '` hook'
+                        message: 'Unexpected use of duplicate Mocha `before` hook'
```
lo1tuma: That message doesn’t seem to make sense for this rule. Shouldn’t be something like "Unnecessary use of Mocha hook for only one test case"?

jfmengels: Absolutely. I simply forgot to add a new message... :sweat_smile: (I'll remove the comment too)

jfmengels:
> If a test requires some clean-up code to run even if the test fails then this must be done in the after hook because the test function exits on the first failing assertion.

True.
Do you think `after` should always be allowed, behind a flag (enabled or not by default?). Same thing for `afterEach`?

lo1tuma:
I would simply always allow to use `after`; `afterEach` could be simply renamed to `after` for a single test case. But maybe should rethink the name of rule. What do you think about `no-unnecessary-hooks`?

jfmengels:
`no-unnecessary-hooks` doesn't fit the bill IMO. If you have a `before` for one single case, the hook is not "unnecessary", it's just that we want to promote an other way of doing it.
`no-setup-for-single-case` / `no-before-for-single-case` would work, but wouldn't catch the `afterEach` meaning. In the direction of `no-common-code-for-single-case` maybe?

I'm fine with the current name, but I don't mind bikeshedding some more if you think it doesn't fit.

lo1tuma:
For `no-hooks-for-single-case` it would be strange to make exceptions by default, that’s why I thought it might be good to find a rule name that doesn’t give the impression that all hooks are forbidden for a single test case. Maybe we can simply replace `no` with `avoid`, which would be less strict?

jfmengels:
Hey, sorry for not replying, I was on vacation :)

> For no-hooks-for-single-case it would be strange to make exceptions by default

I don't think it is that strange, but I agree it could be confusing (but that's why there is documentation). We could forbid `after` by default but push for users to configure it by allowing `after`. I think that is probably the least confusing option. Maybe then have a new `allow` option which takes an array of strings: `['after', 'afterEach']`?

I'd stick with `no-` as that is the convention with ESLint rules, even with similar cases where not everything gets forbidden using options.

lo1tuma:
@jfmengels No problem, I hope you had a great time ;).

Ok, so let’s go with your suggestion and forbid all hooks per default but allow to define exceptions per option.

jfmengels:
> I hope you had a great time ;)

I did :)

> Ok, so let’s go with your suggestion and forbid all hooks per default but allow to define exceptions per option.

:+1:, I'll try to get on this soon. Will fix the other things you mentioned in the same go, like the error message.

jfmengels:
Added an `allow` option, and updated the error message.

docs/rules/no-hooks-for-single-case.md
```diff
@@ -0,0 +1,78 @@
...
+
+    describe('bar', function() {
+      /* ... */
+    });
+});
+```
+
+# Options
+
+This rule supports the following shared configuration options:
```
lo1tuma: `allow` is a rule option not a shared configuration, right?

docs/rules/no-hooks-for-single-case.md
```diff
@@ -0,0 +1,78 @@
...
+
+* `allow`: An array containing the names of hooks to allow. This might be used to allow writing `after` hooks to run clean-up code. Defaults to an empty array.
+
+```json
+{
+    "rules": {
+        "mocha/no-skipped-tests": ["error", {"allow": ["after"]}]
```
lo1tuma: Wrong rule name.

jfmengels:
Oops... good catch. Fixed.

lo1tuma:
Awesome. Thank you!

jfmengels:
My pleasure. That was the last one of the rules that I wrote :smile:
I might take a look at working on others later, there are still some interesting ones in discussion.

PR #97, opened 2016-08-27
Add `no-identical-title` rule (fixes #33)

### LINKED ISSUE #33: Forbid duplicate test names

jfmengels:
Add `no-identical-title` rule (fixes #33)

I decided to name it `no-identical-title` to be "consistent" with `eslint-plugin-ava`'s [similar rule](https://github.com/avajs/eslint-plugin-ava/blob/master/docs/rules/no-identical-title.md). Let me know if you prefer it to be `no-duplicate-title` or something else.

### DIFF
--- a/docs/rules/no-identical-title.md
+++ b/docs/rules/no-identical-title.md
@@ -1,0 +1,45 @@
+# Disallow identical titles (no-identical-title)
+
+Having identical titles for two different tests or test suites may create confusion. For example, when a test with the same title as another test in the same test suite fails, it is harder to know which one failed and thus harder to fix.
+
+## Rule Details
+
+This rule looks at the rule of every test and test suites. It will report when two test suites or two test cases at the same level of a test suite have the same name.
+
+The following patterns are considered warnings:
+
+```js
+describe('foo', function () {
+    it('should do bar', function() {});
+    it('should do bar', function() {}); // Has the same title as the previous test
+
+    describe('baz', function() {
+        // ...
+    });
+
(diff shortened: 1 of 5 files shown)


docs/rules/no-identical-title.md
```diff
@@ -0,0 +1,45 @@
+# Disallow identical titles (no-identical-title)
+
+Having identical titles for two different tests or test suites may create confusion. For example, when a test with the same title as another test in the same test suite fails, it is harder to know which one failed and thus harder to fix.
+
+## Rule Details
+
+This rule looks at the rule of every test and test suites. It will report when two test suites or two test cases at the same level of a test suite have the same name.
```
lo1tuma: I guess `rule` should be `title` or `name`, right?

jfmengels: Absolutely. And `name` should probably be `title` too.

jfmengels:
Fixed. By the way, I'd recommend squashing PRs in order to keep the commit history clean.

lo1tuma:
Merged. Thank you.

I also squashed the commits of this PR but I guess my changelog tool pr-log won’t work with this as it looks specifically for merge commits on `master` in order to create the changelog. Anyway shouldn’t be a problem to create the changelog manually for the next release.

jfmengels:
You're welcome :)

Ah, right. Yeah, should be fine in this case, but that's too bad. Maybe try looking at a different tool to does not rely on merge commits, because a flat commit history is really nice IMO.

PR #101, opened 2016-09-06
Add no-before-after rule

lencioni:
`before()` and `after()` run once before and after all examples. In
practice, this can end up causing confusion regarding the order that
things are run, and test pollution. To improve the readability and
accuracy of your Mocha tests, use `beforeEach()` and `afterEach()`
instead, wherever possible.

I had this rule sitting around in an internal ESLint plugin and figured
I would contribute it to this plugin in case others find it useful.

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -15,3 +15,4 @@
 * [no-hooks-for-single-case](no-hooks-for-single-case.md) - disallow hooks for a single test or test suite
 * [no-top-level-hooks](no-top-level-hooks.md) - disallow top-level hooks
 * [no-identical-title](no-identical-title.md) - disallow identical titles
+* [no-before-after](no-before-after.md) - disallow `before` and `after` in favor of `beforeEach` and `afterEach`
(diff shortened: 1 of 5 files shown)


lencioni:
Looking through the rules here again and I spotted no-hooks. I wonder if this would be better as an option on that rule instead of its own rule. Thoughts?

jfmengels:
This might be a better option on `no-hooks` yeah, maybe a `allow` option.

Just FYI, when I only use `beforeEach`, it's usually to have a fresh clean slate for each test. Now I prefer to do that by calling explicitly a function which gives me the fixtures that I need to use in the test. But this gets harder to do when you use stubs/mocks/spies and that you wish to reset correctly after the test (so an afterEach is nice in that case).

lencioni:
In my rule here, I took some care to ensure that this is only flagged on uses within `context` or `describe`. I noticed that the rules here don't do that. Perhaps that's not important, but if you think it is useful, should we add that type of thing to the rules here?

lo1tuma:
Thanks for this PR. I would also agree that this feature could be implemented as an option of `no-hooks`. It is basically the same option we already have for `no-hooks-for-single-case`.

> In my rule here, I took some care to ensure that this is only flagged on uses within context or describe

Is there a specific reason why top-level/global hooks should be excluded from this rule?

Issue #103, opened 2016-09-25
Rules proposal: one describe per file


alecxe:
The rationale and the use case we have is described here [Enforce one describe per file](http://stackoverflow.com/questions/39436530/enforce-one-describe-per-file).

To summarize - the rule should produce a warning if there is more than one top-level `describe` per file.

Probably, we have too narrow of a use case - feel free to close if you feel there is no need for this kind of rule. Or, may be we can just turn it off by default.

Thanks!

lo1tuma:
Hi @alecxe, thanks for the suggestion. I like the idea and would probably use such a rule as I’m already restrict my test files to only have one top-level describe 😉 .

Just for completeness: there are of course some limitations to static analysis in JavaScript where not all top-level describes can be detected. For example when someone creates test suites dynamically:

```js
function createTestCases(name) {
    describe(name, function () {
        // tests ...
    });
}

createTestCases('foo');
createTestCases('bar');
```

jfmengels:
FYI, this is kind of similar to https://github.com/lo1tuma/eslint-plugin-mocha/issues/83.

lo1tuma:
@jfmengels Good hint. I guess this rule together with a `no-nested-describes` rule would give you the behavior that you can only use one `describe` per file.

jfmengels:
Ah, good point. Fine by me then :+1:

PR #105, opened 2016-09-28
Add 'one-suite-per-file' rule (#103)

alecxe:
This is for the issue #103. Please review.

I think I have not described the motivation for this rule well enough. Please see if you can think of a better way or more reasons to have this rule enabled. Thanks!


docs/rules/one-suite-per-file.md
```diff
@@ -0,0 +1,42 @@
...
+        it('should do bar', function() {});
+    });
+});
+```
+
+There is also possible to configure a custom list of suite names via the second rule configuration option:
+
+```js
+rules: {
+   "mocha/one-suite-per-file": [["describe", "context", "suite", "mysuitename"]]
```
lo1tuma: The severity is missing here, it should be: `["error", ["describe", "context", "suite", "mysuitename"]]`.

lib/rules/one-suite-per-file.js
```diff
@@ -0,0 +1,43 @@
...
+
+module.exports = function (context) {
+    var stack = [],
+        topLevelDescribes = [],
+        suiteNames = context.options[0] ? context.options[0] : defaultSuiteNames;
+
+    return {
+        CallExpression: function (node) {
+            var callee = node.callee;
+            if (callee && callee.name && suiteNames.indexOf(callee.name) > -1) {
```
lo1tuma: Should we also support modifiers as e.g. `descirbe.only`, `describe.skip` etc.?

You could also reuse the [`isDescribe`](https://github.com/lo1tuma/eslint-plugin-mocha/blob/4b7963346efed34a46dc51fa2ad7958b8f3cb2e4/lib/util/ast.js#L13) function.

docs/rules/README.md
```diff
@@ -15,3 +15,4 @@
 * [no-hooks-for-single-case](no-hooks-for-single-case.md) - disallow hooks for a single test or test suite
 * [no-top-level-hooks](no-top-level-hooks.md) - disallow top-level hooks
 * [no-identical-title](no-identical-title.md) - disallow identical titles
+* [one-suite-per-file](one-suite-per-file.md) - disallow multiple top-level suites in a single file
```
alecxe: @lo1tuma absolutely, I like the `max-top-level-suites` - will rename, thanks.

docs/rules/one-suite-per-file.md
```diff
@@ -0,0 +1,42 @@
+# Disallow multiple top-level suites in a single file (one-suite-per-file)
+
+This rule enforces having a single top-level suite in a file.
+
+Multiple `describe` blocks is often a sign that a test should be broken down to multiple files.
+One of the possible problems if having multiple suites is that, if you are, for example, going through tests one by one and focusing them (using `fdescribe`), you might not notice `describe` block(s) that are down there at the bottom of a file which may lead to tests being unintentionally skipped.
```
alecxe: @lo1tuma my bad, too much work with jasmine these days. Will fix.

docs/rules/one-suite-per-file.md
alecxe: @lo1tuma good finding, thanks.

lib/rules/one-suite-per-file.js
alecxe: @lo1tuma yeah, I think we should support all possible suite variations. Will reuse `isDescribe`, thanks.

alecxe: @lo1tuma do you think I can/should safely extend the [`describeAliases`](https://github.com/lo1tuma/eslint-plugin-mocha/blob/4b7963346efed34a46dc51fa2ad7958b8f3cb2e4/lib/util/ast.js#L5) array and include `describe.only`, `describe.skip`, `suite` (`suite.only` and `suite.skip`)?

lo1tuma: I hope so 😄 .  I would rely on our tests here, so if no other test fails after changing `describeAliases` I would say it is safe to change it.

That’s being said, `isDescribe` currently doesn’t support `MemberExpressions` like `describe.skip`, but it could be easily added, just have a look at how it is done in [`isTestCase`](https://github.com/lo1tuma/eslint-plugin-mocha/blob/4b7963346efed34a46dc51fa2ad7958b8f3cb2e4/lib/util/ast.js#L35)

alecxe:
@lo1tuma updated the PR, please review.

lib/rules/max-top-level-suites.js
```diff
@@ -0,0 +1,42 @@
...
+                }
+
+                stack.pop(node);
+            }
+        },
+
+        'Program:exit': function () {
+            if (topLevelDescribes.length > suiteLimit) {
+                context.report({
+                    node: topLevelDescribes[1],
```
lo1tuma: I guess `topLevelDescribes[1]` needs to be changed to `topLevelDescribes[suiteLimit]`, otherwise the rule would always report the second top-level describe node, right?

lib/rules/max-top-level-suites.js
```diff
@@ -1,28 +1,27 @@
...
 
-var defaultSuiteNames = [ 'describe', 'context', 'suite' ];
+var astUtil = require('../util/ast'),
+    defaultSuiteLimit = 1;
 
 module.exports = function (context) {
     var stack = [],
         topLevelDescribes = [],
-        suiteNames = context.options[0] ? context.options[0] : defaultSuiteNames;
+        suiteLimit = context.options[0] ? context.options[0] : defaultSuiteLimit;
```
jfmengels: You could do `context.options[0] || defaultSuiteLimit` (shorter) or `context.options[0] !== undefined ? context.options[0] : defaultSuiteLimit` (maybe a bit more correct, if the user wants 0 suites for some reason)

lib/rules/max-top-level-suites.js
alecxe: @lo1tuma you are right, fixed.

alecxe:
@lo1tuma okay, please see if it is good enough now.

jfmengels:
I still think the options should be an object (https://github.com/lo1tuma/eslint-plugin-mocha/pull/105#pullrequestreview-2312074)

alecxe:
@jfmengels ah, alright, let me fix that, thanks.

alecxe:
@jfmengels okay, done, please see if this is what you've meant. I like the idea of having an object - explicit and readable. Thanks.

lib/rules/max-top-level-suites.js
```diff
@@ -0,0 +1,49 @@
...
+
+module.exports = function (context) {
+    var stack = [],
+        topLevelDescribes = [],
+        suiteLimit = context.options[0];
+
+    if (R.isNil(suiteLimit)) {
+        suiteLimit = defaultSuiteLimit;
+    } else {
+        suiteLimit = suiteLimit.limit;
```
jfmengels: It's possible that users supplied an empty object (or that in the future another option will be available and that the user sets only the value of that one).

The way I usually do this is this way:

```js
var options = context.options[0] || {};
var suiteLimit = options.limit || defaultSuiteLimit;
```

alecxe: @jfmengels okay, good idea, added an empty object handling with a corresponding test. 

By the way, when I write it as:

```js
suiteLimit = defaultSuiteLimit ? R.isNil(options) : options.limit
```

I'm getting a lot of `TypeError: Cannot read property 'loc' of undefined` which I think is a bug in ESLint - so, have to use the explicit if/else in this case.

As far as documenting "options" - I think in the link you've posted "settings" are actually described..but in this case `limit` is a rule option..

Thanks.

jfmengels: > `TypeError: Cannot read property 'loc' of undefined`

No, that sounds like a normal error.

`suiteLimit = defaultSuiteLimit ? R.isNil(options) : options.limit`
`defaultSuiteLimit` is a truthy value, so `suiteLimit` will evaluate to `R.isNil(options)`, meaning `true` or `false`.

When you then compare the number of top-level suites to this boolean, it will evaluate to true if there is at least one. Then, you report an error, but pass undefined as the node (as `topLevelDescribes[true]` doesn't exist), which ESLint then uses to find `loc`. Crash!

tl;dr: You should have written it as `suiteLimit = R.isNil(options.limit) ? defaultSuiteLimit : options.limit)`. I would have done it the way I proposed, but your way should work fine too ;)

jfmengels:
LGTM, thanks @alecxe! :)

alecxe:
Okay, sorry for the rookie mistakes and thanks for being patient!

jfmengels:
No problem (for me at least)! Thanks for your work! :)

PR #109, opened 2016-10-06
Add no-nested-tests rule

### LINKED ISSUE #108: Rule proposal: no-nested-its
Today I accidentally wrote some tests like this:

```
describe("a", function() {
  it("b", function() {
    it("c1", function() {
       expect(...);
    });

    it("c2", function() {
       expect(...);
    });
  });
});
```

instead of like this:

```
describe("a", function() {
  describe("b", function() { // <-----------
    it("c1", function() {
       expect(...);
    });

    it("c2", function() {
       expect(...);
    });
  });
});
```

This caused mocha to not run tests "c1" or "c2" at all, since they were accidentally declared inside an `it` block instead of a `describe` block.

I only noticed because my code coverage tool said it was missing certain execution branches that I could have sworn I had written tests for. A linter setting to point out this subtle mistake would be helpful IMO.

lo1tuma:
fixes #108

### DIFF
--- a/lib/rules/no-nested-tests.js
+++ b/lib/rules/no-nested-tests.js
@@ -1,0 +1,38 @@
+'use strict';
+
+var astUtils = require('../util/ast');
+
+module.exports = function noNestedTests(context) {
+    var testNestingLevel = 0;
+
+    function report(callExpression, isTestCase) {
+        var message = isTestCase ? 'Unexpected test nested within another test.' :
+            'Unexpected suite nested within a test.';
+
+        context.report({
+            message: message,
+            node: callExpression.callee
+        });
+    }
+
+    return {
+        CallExpression: function (node) {
+            var isTestCase = astUtils.isTestCase(node),
+                isDescribe = astUtils.isDescribe(node);
+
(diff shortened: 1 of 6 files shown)


lib/rules/no-nested-tests.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+var astUtils = require('../util/ast');
+
+module.exports = function noNestedTests(context) {
+    var testNestingLevel = 0;
+
+    function report(callExpression) {
+        var message = astUtils.isTestCase(callExpression) ? 'Unexpected test nested within another test.' :
```
jfmengels: *slight* boost, you call have an argument `isTestCase` for this function, as you have already computed in the calling function.

lo1tuma: Good hint 👍 
Done 😉

lo1tuma:
> do you update the changelog when releasing?

Yes. When doing a release I run `npm run changelog <version-number>` to generate the changelog based on the titles of closed PRs since the last release.

PR #112, opened 2016-10-26
Make no-mocha-arrows rule fixable

rhysd:
## Summary

I make `no-mocha-arrows` rule fixable with `--fix` option of `eslint` command.
## Example
- Before

``` js
it("does something", () => { assert.ok(true); });
it("does something", () => assert.ok(true));
```
- After

``` js
it("does something", function () { assert.ok(true); });
it("does something", function () { return assert.ok(true); });
```

EDIT: Translation result of single expression body

### DIFF
--- a/lib/rules/no-mocha-arrows.js
+++ b/lib/rules/no-mocha-arrows.js
@@ -41,6 +41,38 @@ module.exports = function (context) {
         return !R.find(R.propEq('name', name), scope.variables);
     }
 
+    function fixArrowFunction(fixer, fn) {
+        var sourceCode = context.getSourceCode(),
+            paramsLeftParen = sourceCode.getFirstToken(fn),
+            paramsRightParen = sourceCode.getTokenBefore(sourceCode.getTokenBefore(fn.body)),
+            paramsFullText =
+                sourceCode.text.slice(paramsLeftParen.range[0], paramsRightParen.range[1]),
+            functionKeyword = 'function ',
+            bodyText;
+
+        if (fn.async) {
+            // When 'async' specified, take care about the keyword.
+            functionKeyword = 'async function';
+            // Strip 'async (...)' to ' (...)'
+            paramsFullText = paramsFullText.slice(5);
+        }
+
+        if (fn.body.type === 'BlockStatement') {
+            // When it((...) => { ... }),
+            // simply replace '(...) => ' with 'function () '
+            return fixer.replaceTextRange(
+                [ fn.start, fn.body.start ],
+                functionKeyword + paramsFullText + ' '
+            );
+        }
+
+        bodyText = sourceCode.text.slice(fn.body.range[0], fn.body.range[1]);
+        return fixer.replaceTextRange(
+            [ fn.start, fn.end ],
+            functionKeyword + paramsFullText + ' { return ' + bodyText + '; }'
+        );
+    }
+
     return {
         CallExpression: function (node) {
             var name = getCalleeName(node.callee),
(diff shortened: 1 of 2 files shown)


rhysd:
Travis CI workers don't start... Could you restart them?

test/rules/no-mocha-arrows.js
```diff
@@ -23,28 +23,39 @@ ruleTester.run('no-mocha-arrows', rules['no-mocha-arrows'], {
...
-            errors: [ { message: expectedErrorMessage, column: 1, line: 1 } ]
+            errors: [ { message: expectedErrorMessage, column: 1, line: 1 } ],
+            output: 'it(function () { assert(something, false); })'
         },
         {
             code: 'it(() => assert(something, false))',
             parserOptions: { ecmaVersion: 6 },
-            errors: [ { message: expectedErrorMessage, column: 1, line: 1 } ]
+            errors: [ { message: expectedErrorMessage, column: 1, line: 1 } ],
+            output: 'it(function () { assert(something, false); })'
```
jfmengels: The output should be
`it(function () { return assert(something, false); })`

Imagine the user wrote a test for function that returns a Promise, and they just want to check that it resolves.

```js
it(() => lib.resolves());
```

If it's changed to

```js
it(() => {
  lib.resolves();
});
```

then the test will passed even if the Promise rejects. It's much safer to change it to

```js
it(function() {
  return lib.resolves();
});
```

rhysd: Ah, you're correct.  I'll fix this.

rhysd:
My TODO:

- [ ] Add test cases for generator function, async function and `done` callback
- [ ] Missing `return` statement

rhysd:
@jfmengels 

I fixed all the points. Could you review additional patches?

lib/rules/no-mocha-arrows.js
```diff
@@ -50,7 +50,41 @@ module.exports = function (context) {
...
+                                    paramsFullText =
+                                        sourceCode.text.slice(paramsLeftParen.range[0], paramsRightParen.range[1]),
+                                    functionKeyword = 'function ',
+                                    bodyText;
+
+                                if (fnArg.async) {
+                                    // When 'async' specified, take care about the keyword.
+                                    functionKeyword = 'async ' + functionKeyword;
+                                    // Strip 'async (...)' to '(...)'
+                                    paramsFullText = paramsFullText.slice(6);
```
jfmengels: Both `async () => {}` and `async() => {}` are actually valid. This would create broken code if you find the latter form. (add a test for that obviously :) )

test/rules/no-mocha-arrows.js
```diff
@@ -16,35 +16,73 @@ ruleTester.run('no-mocha-arrows', rules['no-mocha-arrows'], {
...
-        'it.only()'
+        'it.only()',
+        'it(function(done) { assert(something, false); done(); })',
+        {
+            code: 'it(function*() { assert(something, false) })',
+            parserOptions: { ecmaVersion: 6 }
+        },
+        {
+            code: 'it(async function () { assert(something, false) })',
+            parserOptions: { ecmaVersion: 2017 }
```
jfmengels: Pretty sure you can do
```js
ruleTester = new RuleTester({
  parserOptions: {ecmaVersion: 2017}
});
```
and remove all the parserOptions override in individual test cases.

test/rules/no-mocha-arrows.js
```diff
@@ -16,35 +16,73 @@ ruleTester.run('no-mocha-arrows', rules['no-mocha-arrows'], {
...
+        }
     ],
 
     invalid: [
         {
             code: 'it(() => { assert(something, false); })',
             parserOptions: { ecmaVersion: 6 },
-            errors: [ { message: expectedErrorMessage, column: 1, line: 1 } ]
+            errors: [ { message: expectedErrorMessage, column: 1, line: 1 } ],
```
jfmengels: I usually extract the common errors into a variable and just put it everywhere. Removes a lot of duplicated code. but @lo1tuma may not have the same preference.

```js
const errors =  [ { message: 'Do not pass arrow functions to it()', column: 1, line: 1 } ];

 ruleTester.run('no-mocha-arrows', rules['no-mocha-arrows'], {
 // ...
  {
    code: '...',
    errors: errors
  }
});
```

lib/rules/no-mocha-arrows.js
rhysd: Hmm... It looks nice point. I need to improve code transform on async function.

test/rules/no-mocha-arrows.js
rhysd: Oh, that's much better.  I don't know eslint toolchain very much. Thanks.

rhysd:
Thank you @jfmengels for nice points. I'll fix them.

lib/rules/no-mocha-arrows.js
rhysd: Let me confirm one point. When `async()`, should we transform it to `async function()`? Or is it OK to transform it to `async function ()`?

rhysd:
@jfmengels 

I added extra 4 commits. Each commit correspond to your review comment. Could you review them?

test/rules/no-mocha-arrows.js
```diff
@@ -2,49 +2,90 @@
...
+        {
+            code: 'it("should be false", () => {\n assert(something, false);\n})',
+            errors: errors,
+            output: 'it("should be false", function () {\n assert(something, false);\n})'
+        }
+    ]
+
+});
+
+es2017RuleTester.run('no-mocha-arrows', rules['no-mocha-arrows'], {
```
jfmengels: Would have been simpler (and less confusing) to have only one `ruleTester.run` call with the es6 settings, and just override the parser settings where needed (in the following cases).

rhysd: I feel it's enough to use test runner using es2016 parser for every tests.  I'll make one test runner for tests.

rhysd:
@jfmengels I fixed the point :smile:

lo1tuma:
LGTM as well 👍 .

Thanks everyone.

PR #118, opened 2016-12-13
Fix no-mocha-arrow fixer

jfmengels:
When a test implementation has parameters, the end result would have missing parens.

```js
it(done => {});
// -->
it(function done {});
```

This fixes it. I could not keep a consistent spacing for `function () {}` and `function() {}`, so I modified the current results, but when fixing, that will probably be taken care of the other rules' fixers anyway.

cc @rhysd if you feel like reviewing


rhysd:
Ah, you're correct. Thank you for the fix.

PR #126, opened 2017-03-15
Add settings to enable support for mocha.parallel

stefanbuck:
This PR adds a new shared settings value which accepts an array of additional suite names, for example:

`
  "settings": {
      "mocha/additionalSuiteNames': [ 'foo', 'bar' ],
  }
`

This is useful when working with modules like [mocha.parallel](https://github.com/danielstjules/mocha.parallel)

@lo1tuma let me know what you think and I'll adapt it for the following rules as well.

- [x] `no-top-level-hooks`
- [x] `max-top-level-suites`
- [x] `no-hooks-for-single-case`
- [x] `no-identical-title`
- [x] `no-nested-tests`
- [x] `no-sibling-hooks`

### DIFF
--- a/test/rules/no-top-level-hooks.js
+++ b/test/rules/no-top-level-hooks.js
@@ -21,9 +21,21 @@ ruleTester.run('no-top-level-hooks', rules['no-top-level-hooks'], {
         'after.foo()',
         'beforeEach.foo()',
         'afterEach.foo()',
-        'var before = 2; before + 3;'
+        'var before = 2; before + 3;',
+        {
+            code: 'foo(function() { before(function() {}); });',
+            settings: {
+                'mocha/additionalSuiteNames': [ 'foo' ]
+            }
+        }, {
+            code: 'foo(function() { before(function() {}); });',
+            settings: {
+                mocha: {
+                   additionalSuiteNames: [ 'foo' ]
+               }
+           }
+         }
     ],
-
     invalid: [
         {
             code: 'before(function() {});',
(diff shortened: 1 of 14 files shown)


test/rules/no-top-level-hooks.js
```diff
@@ -22,7 +22,24 @@ ruleTester.run('no-top-level-hooks', rules['no-top-level-hooks'], {
         'beforeEach.foo()',
         'afterEach.foo()',
         'var before = 2; before + 3;'
-    ],
+    ].reduce(function (memo, test) {
+        memo.push(test);
+        memo.push({
+            code: test,
```
stefanbuck: I just realized that the test isn't correct implement, because the code doesn't contain `parallel`. Will fix it tomorrow

lo1tuma:
Thanks for the PR.

Actually I was thinking about a more generic approach. I would prefer a setting which accepts an array of additional suite names, for example:

```
"settings": { "mocha/additionalSuiteNames": [ "parallel", "xparallel", "myOtherMochaExtensionCustomSuiteName" ] } }
```

This way we would be agnostic to any specific mocha extension and we don’t have to add a separate setting for each custom suite name. What do you think?

stefanbuck:
Yeah, make sense. I've updated the PR as you suggested. Let me know if it's okay for you and I'll tweak the other rules as well.

stefanbuck:
@lo1tuma please have another look. I'm undecided if the valid test cases really make sense, since the suite names is optional.

lo1tuma:
@stefanbuck I think those tests make sense, if someone breaks the feature of `addtionalSuteNames` those tests would fail.

Issue #131, opened 2017-06-03
Support async functions


jawadst:
**Proposal:** Support [async functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function)

**1) Update `no-synchronous-tests`** 
It would be great to update the `no-synchronous-tests` rule to support async functions.

The updated rule would pass if one of the following is true:

- There is a `done` parameter
- There is a return
- The function provided to mocha is async

Example:
```
it('test with an async function', async function () {
   ...
});
```

**2) Rule: Force async functions**
Add a rule that prevents anything else than async functions given to Mocha (no done and no sync test).

lo1tuma:
Sounds good to me.

jawadst:
@lo1tuma 
I have pushed a PR for these changes. Lmk if you want me to change anything before you can merge.

lo1tuma:
I wonder if we should make `no-synchronous-tests` configurable instead of creating a new rule. We could introduce an option which allows to define the preferred async test method. This way we could also enforce to always use promises or callbacks. 

Example:

```
"no-synchronous-tests" : [ "error", { allowed: [ "async"]}] 
```

The default would be `["async", "promise", "callback"] `. 

What do you think?

jawadst:
I think that's a good idea because it allows to (also) only accept promises or callbacks which is a best practice, I think, to ensuri that tests are homogeneous.

I'll update the code for that. One PR is enough then? (as I'll only be modifying the "no-synchronous-tests" rule and not adding a new one).

lo1tuma:
> One PR is enough then?

Yes, would be ok for me 😉.

jawadst:
New PR: https://github.com/lo1tuma/eslint-plugin-mocha/pull/138

lo1tuma:
Releases in 4.11.0

PR #136, opened 2017-06-12
only-async-functions + Allowing async functions in no-synchronous-tests

jawadst:
See https://github.com/lo1tuma/eslint-plugin-mocha/issues/131


lo1tuma:
Thanks for the PR. Could you make a separate PR for the new rule?

PR #137, opened 2017-06-12
don't drop support for eslint 3.x

modosc:
unless i'm missing something there's no reason to drop this?


lo1tuma:
Fix released in v4.10.1.

PR #138, opened 2017-06-14
Added support for async functions in no-synchronous-tests

jawadst:
One note: the updated rule throws errors when the provided options do not match a certain format.
eslint itself usually does not test those exception cases (it ignores them for code coverage: https://github.com/eslint/eslint/blob/master/lib/rules/new-cap.js#L38) and I wasn't sure how to test an exception with the RuleTester so I ignored those cases as well.

### DIFF
--- a/lib/rules/no-synchronous-tests.js
+++ b/lib/rules/no-synchronous-tests.js
@@ -1,9 +1,8 @@
 'use strict';
 
-var R = require('ramda');
-
-module.exports = function (context) {
-    var possibleAsyncFunctionNames = [
+var R = require('ramda'),
+    defaultAllowedAsyncMethods = [ 'async', 'callback', 'promise' ],
+    possibleAsyncFunctionNames = [
         'it',
         'it.only',
         'test',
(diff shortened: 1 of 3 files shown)


lo1tuma:
To validate `options` you can provide a JSON schema which ESLint will take to check if the passed options match agains the schema. So I guess we could just define a [schema](http://eslint.org/docs/developer-guide/working-with-rules#options-schemas) here instead of implementing our own validation logic.

lib/rules/no-synchronous-tests.js
```diff
@@ -16,54 +15,105 @@ module.exports = function (context) {
         'afterEach'
     ];
 
-    function getCalleeName(callee) {
-        if (callee.type === 'MemberExpression') {
-             return callee.object.name + '.' + callee.property.name;
-        }
-
-        return callee.name;
+function getCalleeName(callee) {
```
lo1tuma: I know this function was already there before but it would be nice to re-use the [`getNodeName`](https://github.com/lo1tuma/eslint-plugin-mocha/blob/0e54998108f86bae7ddb029aae06be6d128fc652/lib/util/ast.js#L17) method from our AST utils.

💅

lib/rules/no-synchronous-tests.js
```diff
@@ -16,54 +15,105 @@ module.exports = function (context) {
...
+        && typeof returnStatement !== 'undefined';
+}
 
-        if (bodyStatement.type === 'BlockStatement') {
-            returnStatement = findPromiseReturnStatement(functionExpression.body.body);
-        } else if (bodyStatement.type === 'CallExpression') {
-            //  allow arrow statements calling a promise with implicit return.
-            returnStatement = bodyStatement;
-        }
+function getAllowedAsyncMethocsFromOptions(options) {
```
lo1tuma: Typo? I guess it should be `getAllowedAsyncMethodsFromOptions`.

jawadst:
I have implemented the requested changes. I am using isTestCase/isHookIdentifier from ast utils and got rid of the list of identifiers for tests/hooks in no-synchronous-tests.

PR #146, opened 2017-11-05
Adds rule "padding-between-test-functions"

### LINKED ISSUE #145: Rule proposal: spaces between test blocks
I tried to get some combination of existing rules from eslint to describe this style, but couldn't quite do it without adversely affecting the code inside the test block. 

I propose a rule that affects expression statements with a name of one of the mocha test functions (describe, it, before/after/etc) and enforces spacing before and after the test with options for always or never.

Example of bad code with option 'always':

```
describe('some describe', () => {
  it('should really have spacing', () => {})
})
```

Example of bad code with option 'never':
```
describe('some describe', () => {

  it('shouldnt have spacing', () => {})

})
```

I have this written and was just going to include it in a custom plugin - but then I saw this project existed so I figured I'd share it.

tswaters:
Fixes #145 

--

As an aside, I had a heck of a time getting my plugin, as-written, to fit in the eslint ruleset of this project:

- complexity of 4 meant i needed to abuse nested functions a bit... the function really needs complexity of 5 to work properly (exit early if statement; each option has different reports/fixers for padding above/below)

- vars-on-top means I needed to split out the calls to `isPaddingBetweenTokens` into individual functions... it would normally exit early based on results of var declarations, and set up `hasTop/BottomPadding` afterwards.

It's passing the test suite / linting rules for the project, but I fear it has gotten more complex and harder to follow as a result.

### DIFF
--- a/docs/rules/padding-between-test-functions.md
+++ b/docs/rules/padding-between-test-functions.md
@@ -1,0 +1,72 @@
+# Padding Between Test Functions (padding-between-test-functions)
+
+This is a stylistic rule that enforces or disallows spaces between test functions.
+
+**Fixable:** Problems detected by this rule are automatically fixable using the `--fix` flag on the command line.
+
+## Rule Details
+
+This rule looks for every call to  `describe`, `context`, `it`, `specify`, `before`,  `after`, `beforeEach`, `afterEach`, `suite`, `test`, `suiteSetup`, `suiteTeardown`, `setup`, `teardown` and will ensure consistent spacing between the nearest token.
+
+## Options
+
+This rule has one option:
+
+* `"always"` (default) requires empty lines at the beginning and ending of test functions
+* `"never"` disallows empty lines at the beginning and ending of test functions
+
+### always
+
+Examples of **incorrect** code for this rule with the default `"always"` option:
+
+```js
+describe(function() {
+    beforeEach(function(){});
+    afterEach(function(){});
+    it('', function() {});
+});
+```
+
+Examples of **correct** code for this rule with the default `"always"` option:
+
+```js
+describe(function() {
+
+    beforeEach(function(){});
+
+    afterEach(function(){});
+
+    it('', function() {});
+
+});
+```
+
+### never
+
+Examples of **incorrect** code for this rule with the `"never"` option:
+
+```js
+describe(function() {
+
+    beforeEach(function(){});
+
+    afterEach(function(){});
+
+    it('', function() {});
+
+});
+```
+
+Examples of **correct** code for this rule with the `"never"` option:
+
+```js
+describe(function() {
+    beforeEach(function(){});
+    afterEach(function(){});
+    it('', function() {});
+});
+```
+
+## When Not To Use It
+
+You can turn this rule off if you are not concerned with the consistency of padding between test functions.
--- a/docs/rules/padding-between-test-functions.md
+++ b/docs/rules/padding-between-test-functions.md
@@ -1,12 +1,12 @@
 # Padding Between Test Functions (padding-between-test-functions)
 
-This is a stylistic rule that enforces or disallows spaces between test functions.
+This is a stylistic rule that enforces consistent empty line padding before and after mocha functions.
 
 **Fixable:** Problems detected by this rule are automatically fixable using the `--fix` flag on the command line.
 
 ## Rule Details
 
-This rule looks for every call to  `describe`, `context`, `it`, `specify`, `before`,  `after`, `beforeEach`, `afterEach`, `suite`, `test`, `suiteSetup`, `suiteTeardown`, `setup`, `teardown` and will ensure consistent spacing between the nearest token.
+This rule looks for every call to  `describe`, `context`, `it`, `specify`, `before`,  `after`, `beforeEach`, `afterEach`, `suite`, `test`, `suiteSetup`, `suiteTeardown`, `setup`, `teardown` and will ensure consistent spacing between it and the nearest token.
 
 ## Options
 
--- a/lib/rules/padding-between-test-functions.js
+++ b/lib/rules/padding-between-test-functions.js
@@ -13,6 +13,90 @@ function isPaddingBetweenTokens(first, second) {
     return second.loc.start.line - first.loc.end.line >= 2;
 }
 
+/**
+ * Returns a fixer function for replacing multiple new lines with a single one.
+ * @param {Range} first first range to work against
+ * @param {Range} second second range to work against
+ * @returns {function} fixer function.
+ */
+function removeEmptyLinesBetweenTokens(first, second) {
+    return function (fixer) {
+        return fixer.replaceTextRange([ first.range[1], second.range[0] ], '\n');
+    };
+}
+
+/**
+ * Returns a function used to check top-padding.
+ * @param {*} sourceCode sourceCode from eslint plugin api.
+ * @param {string} opt always or never.
+ * @param {string} message message to use for error reporting.
+ * @returns {function} reporting function that returns array of errors.
+ */
+function getTopCheker(sourceCode, opt, message) {
+    return function (node) {
+        var firstToken = sourceCode.getFirstToken(node),
+        tokenBeforeFirst = sourceCode.getTokenBefore(firstToken, { includeComments: true }),
+        blockHasTopPadding = isPaddingBetweenTokens(tokenBeforeFirst, firstToken),
+        errors = [];
+
+        if (blockHasTopPadding && opt === 'never') {
+            errors.push({
+                node: node,
+                message: message,
+                fix: removeEmptyLinesBetweenTokens(tokenBeforeFirst, firstToken)
+            });
+        }
+
+        if (!blockHasTopPadding && opt === 'always') {
+            errors.push({
+                node: node,
+                message: message,
+                fix: function (fixer) {
+                    return fixer.insertTextBefore(firstToken, '\n');
+                }
+            });
+        }
+
+        return errors;
+    };
+}
+
+/**
+ * Returns a function used to check top-padding.
+ * @param {*} sourceCode sourceCode from eslint plugin api.
+ * @param {string} opt always or never.
+ * @param {string} message message to use for error reporting.
+ * @returns {function} reporting function that returns array of errors.
+ */
+function getBottomCheker(sourceCode, opt, message) {
+    return function (node) {
+        var lastToken = sourceCode.getLastToken(node),
+        tokenAfterLast = sourceCode.getTokenAfter(lastToken, { includeComments: true }),
+        blockHasBottomPadding = isPaddingBetweenTokens(lastToken, tokenAfterLast),
+        errors = [];
+
+        if (blockHasBottomPadding && opt === 'never') {
+            errors.push({
+                node: node,
+                message: message,
+                fix: removeEmptyLinesBetweenTokens(lastToken, tokenAfterLast)
+            });
+        }
+
+        if (!blockHasBottomPadding && opt === 'always') {
+            errors.push({
+                node: node,
+                message: message,
+                fix: function (fixer) {
+                    return fixer.insertTextAfter(lastToken, '\n');
+                }
+            });
+        }
+
+        return errors;
+    };
+}
+
 module.exports = {
     meta: {
         docs: {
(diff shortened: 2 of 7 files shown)


lib/rules/padding-between-test-functions.js
```diff
@@ -0,0 +1,142 @@
...
+                            }
+                        });
+                    }
+                    if (blockHasBottomPadding) {
+                        context.report({
+                            node: node,
+                            message: message,
+                            fix: function (fixer) {
+                                return fixer.replaceTextRange([ lastToken.range[1], tokenAfterLast.range[0] ], '\n');
+                            }
```
lo1tuma: Looks like there is a little bit of code duplication in those fixer functions. What do you think about extracting this to a function that creates line removal fixer functions for the given tokens?

Example:

```js
function removeEmptyLinesBetweenTokens(first, second) {
    return function (fixer) {
        return fixer.replaceTextRange([ first.range[1], second.range[0] ], '\n');
    };
}
```

test/rules/padding-between-test-function.js
```diff
@@ -0,0 +1,49 @@
...
+            errors: [ { message: NEVER_MESSAGE }, { message: NEVER_MESSAGE } ],
+            output: 'describe(function(){\nit("", function () {})\n})'
+        },
+        {
+            code: 'describe(function(){\nit("", function () {})\n})',
+            options: [ 'always' ],
+            errors: [ { message: ALWAYS_MESSAGE }, { message: ALWAYS_MESSAGE } ],
+            output: 'describe(function(){\n\nit("", function () {})\n\n})'
+        }
+    ]
```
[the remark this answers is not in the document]
lo1tuma: There is currently also no test case for multiple empty lines between test functions.

tswaters:
Would you like additional commits or an amended commit with the changed?

lo1tuma:
@tswaters Additional commits are ok 😉 .

lib/rules/padding-between-test-functions.js
```diff
@@ -0,0 +1,142 @@
...
+                function shouldReturn() {
+                    return (
+                        !sourceCode.getTokenBefore(node) ||
+                        !sourceCode.getTokenAfter(node) ||
+                        !isTestFunction(node)
+                    );
+                }
+
+                function always() {
+                    var blockHasTopPadding = isPaddingBetweenTokens(tokenBeforeFirst, firstToken),
```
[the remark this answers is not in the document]
tswaters: In my eyes, with never it should be:

```js
[1,2,3].forEach(function () {
  it("foo", function () {});
});
```

...and with always:
```js
[1,2,3].forEach(function () {

  it("foo", function () {});

});
```

The code currently handles this case.  I'll add a test case to ensure this is the case.

lib/rules/padding-between-test-functions.js
```diff
@@ -0,0 +1,142 @@
...
+            ExpressionStatement: function (node) {
+                var firstToken = sourceCode.getFirstToken(node),
+                    tokenBeforeFirst = sourceCode.getTokenBefore(firstToken, { includeComments: true }),
+                    lastToken = sourceCode.getLastToken(node),
+                    tokenAfterLast = sourceCode.getTokenAfter(lastToken, { includeComments: true });
+
+                function shouldReturn() {
+                    return (
+                        !sourceCode.getTokenBefore(node) ||
+                        !sourceCode.getTokenAfter(node) ||
```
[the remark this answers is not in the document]
tswaters: This may be indicative of the complexity of this function - but it'll check both before and after a given mocha function.

So in this case, it checks above describe finds no tokens & aborts; checks below, finds `somethingElse` without padding and reports (if options match)

I've split this out a bit into `topChecker` and `bottomChecker` -- this helps the complexity of the function and hopefully makes things a bit clearer.

I've added test cases to check this particular case.

lib/rules/padding-between-test-functions.js
tswaters: Good call, I'll do that.

lib/rules/padding-between-test-functions.js
```diff
@@ -0,0 +1,142 @@
...
+    'suite',
+    'test',
+    'suiteSetup',
+    'suiteTeardown',
+    'setup',
+    'teardown'
+];
+
+/**
+ * Checks if a node is a test function expression, or object (e.g., it() or it.only())
```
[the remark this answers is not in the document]
tswaters: Good call, I didn't realize those utility functions existed.  Probably quite a bit cleaner than my `isTestFunction` routine. 

However, because it is split out into 3 different functions, I'm running into complexity problems again due to all the or branches...

I'm going to add the missing hooks, and another util, `isTestFunction` that returns true for any of them.

docs/rules/padding-between-test-functions.md
```diff
@@ -0,0 +1,72 @@
+# Padding Between Test Functions (padding-between-test-functions)
+
+This is a stylistic rule that enforces or disallows spaces between test functions.
```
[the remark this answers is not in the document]
tswaters: I'll update this a bit to be more clear

test/rules/padding-between-test-function.js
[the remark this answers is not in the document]
tswaters: Hmmm.... yes, multiple empty lines.  The code doesn't handle this case currently.  While it is possible to add an option for, say, number of spaces allowed before/after test functions - I believe this might trample on another eslint rule -- `no-multiple-empty-lines`.  I think it makes the most sense to let the user decide what is best here by setting that rule.

That said, I have noticed a few other problems in adding more tests like you suggested -- the trailing newline doesn't account for semicolons and will dutifully add a lineline between the end of statement `)` and the semicolon.  Working on this now.

tswaters:
Hm... found a problem which I think is going to take some time to figure out.  

I was hoping to get this into a stable state today, but it seems this is unlikely.  I'll commit what I have - including some broken test cases - and will come back to it hopefully next weekend.

The problem I'm having is extra errors are being reported (and fixes applied) when there are multiple test blocks that right next to each other.  i.e., for `always`:

```
it()
it()
```

An error is reported for padding after the first one, and padding before the second one, so twice as many errors as should be - and also the fix is being applied for each error so it winds up being:

```
it()


it()
```

lib/rules/padding-between-test-functions.js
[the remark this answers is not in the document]
lo1tuma: Ok, so in this case I think the rule name might be a bit misleading. The padding is not only enforced **between** (two) test functions but rather **around** every single test function.

test/rules/padding-between-test-function.js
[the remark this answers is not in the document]
lo1tuma: Huh? Isn’t this condition `second.loc.start.line - first.loc.end.line >= 2;` allowing more than one empty line? IMHO this is also the right thing to do, I just wanted to see a test that the rule also works correctly when the padding is more than one empty line.

lo1tuma:
Good catch. I think the best would be to only report the first detected problem and skip the second one. Maybe you could implement this by collecting all possible rule violations first and once you have all then you can filter out the ones with duplicated source code ranges. After that you can report all remaining violations in the `"Program:exit"` listener.

lib/rules/padding-between-test-functions.js
```diff
@@ -0,0 +1,142 @@
...
+}
+
+/**
+ * Returns a function used to check top-padding.
+ * @param {*} sourceCode sourceCode from eslint plugin api.
+ * @param {string} opt always or never.
+ * @param {string} message message to use for error reporting.
+ * @returns {function} reporting function that returns array of errors.
+ */
+function getTopCheker(sourceCode, opt, message) {
```
lo1tuma: Looks like a typo, I guess it should be getTopChe**c**ker.

tswaters:
Just realized I had a typo in that commit message, should be "typo".  leaving it there for ironic purposes.

I've fixed up the problems.  I see ramda is a dependency here, so I tried my best to use it; I've never used it before, so please let me know if there is anything wrong with my usage.

I've added more assertions on the errors to include start/end positions.  The context reporting has also been updated to correctly reflect the start/end line/col where the white space should/should not be.

lib/rules/padding-between-test-functions.js
```diff
@@ -0,0 +1,166 @@
...
+            },
+            ExpressionStatement: function (node) {
+                var tokenBefore = sourceCode.getTokenBefore(node),
+                    tokenAfter = sourceCode.getTokenAfter(node);
+
+                if (!tokenBefore || !tokenAfter || !astUtil.isTestFunction(node.expression)) {
+                    return;
+                }
+
+                errors = R.reduce(
```
lo1tuma: Is the usage of `R.reduce` really necessary here? I think `errors = R.concat(errors, [ topChecker(node), bottomChecker(node) ])` should be sufficient.

lib/util/ast.js
```diff
@@ -5,7 +5,8 @@
 var describeAliases = [ 'describe', 'xdescribe', 'describe.only', 'describe.skip',
                         'context', 'xcontext', 'context.only', 'context.skip',
                         'suite', 'xsuite', 'suite.only', 'suite.skip' ],
-    hooks = [ 'before', 'after', 'beforeEach', 'afterEach' ],
+    hooks = [ 'before', 'after', 'beforeEach', 'afterEach', 'setup', 'tearDown',
+              'suiteSetup', 'suiteTeardown' ],
```
lo1tuma: 👍

test/rules/padding-between-test-function.js
```diff
@@ -0,0 +1,258 @@
...
+            }, {
+                message: ALWAYS_MESSAGE,
+                line: 4,
+                column: 40,
+                endLine: 5,
+                endColumn: 1
+            } ],
+            output: [
+                'describe(function(){',
+                'var someVar=null;',
```
lo1tuma: Is this really the expected behavior or should there be another blank line between `describe` and `var`?

lib/rules/padding-between-test-functions.js
tswaters: That's what I tried initially but `R.concat` has arity of 2 - it doesn't care about subsequent parameters.  When passing an array like that it turns it into

```
[
  [
    {top-error},
    {bottom-error},
  ],
  //...etc...
]
```

I stumped across that pattern here: https://stackoverflow.com/a/34376699.

test/rules/padding-between-test-function.js
lo1tuma: I think I got it now. I was confused for a moment and thought the padding of the `describe` block should also affect the inner part, but that doesn’t make sense after a second thought.

lib/rules/padding-between-test-functions.js
lo1tuma: Ah, you are right, that’s because `topChecker` and `bottomChecker` returns an array as well.

docs/rules/padding-between-test-functions.md
```diff
@@ -0,0 +1,72 @@
+# Padding Between Test Functions (padding-between-test-functions)
```
lo1tuma: So what do you think about the rule name? As I understand it the rule enforces the padding around each test function and not specifically between two test functions. So maybe a better name would be `padding-around-test-functions`?

jfmengels:
This is a great idea. I was actually thinking of writing this rule myself.

One thing that bothers me though: The way I write my tests is like this, so the same thing as with the `always` option, but without the empty line at the start of `describe`.

Example:
```js
describe('some describe', () => {
  it('shouldnt have spacing', () => {
  });

  it('shouldnt have spacing', () => {
  });

  it('shouldnt have spacing', () => {
  });

  describe('some nested describe', () => {
    it('shouldnt have spacing', () => {
    });
  });
})
```

I have also turned on the core ESLint [padded-blocks](https://eslint.org/docs/rules/padded-blocks) rule, which forbids an empty line at the start of the function. When autofixing, having the two rules would make ESLint go back and forth here.

I would suggest to either:
- Change the rule to enforce the style I just described. If chosen, the rule might be better called `spacing-between-test-functions`
- Add an extra option dealing that supports this style (and also warn users about the potential conflict with `padded-blocks`)

Tell me what you think. And thanks @tswaters for your work on this! :)

tswaters:
@jfmengels how about an option for `padded-blocks`  with the same options?  This would inspect any of the `describeAliases` and enforce or forbid padding similar to how `padded-blocks` does it.

AFAIK, eslint rules can't read other config options so a user of this rule would need to specify the same, if desired... otherwise, like you say, eslint would toggle between the two.

I'd need to think about how that would work and I'm not sure I'll have time in the immediate future to work on this.  I've given you both contributor access to my fork if you want a kick at the can 😄

mschilde:
Is there any chance that this PR gets merged anytime soon? How can I help?

trollepierre:
Hi guys, what's up with this PR ?

pkuczynski:
Yeah, I would love to see this PR completed as @jfmengels described. @lo1tuma what's missing so I can maybe help?

tswaters:
Hi sorry for not responding here.

If anyone is interested here let me know and I can grant access to my fork.  Another option would be to fork my fork and go from there.  

The changes are all on master, from a while back --so you're probably going to want to rebase that commit against origin.

I don't have any time to work on this so I'm going to  close this PR, and someone else can take up the mantle.

PR #147, opened 2018-02-02
Adds rule "no setup in describe"

donabrams:
Setup for test cases in mocha should be done in `before`, `beforeEach`, or `it` blocks. Unfortunately there is nothing stopping you from doing setup directly inside a `describe` block and it's not immediately obvious setup should be inside a `before` or `beforeEach`.

This rule catches many cases where setup is accidentally done directly in a describe block and promotes the use of the hooks `before` and `beforeEach`.

This rule's code is a little more complex than I'd like because it does what it can to minimize processing time outside a describe block (`Array.length` check only).

This rule should never be activated by default because it blocks the use of dynamically generated tests.


.eslintrc
```diff
@@ -103,7 +103,7 @@
         "comma-dangle": [ 2, "never" ],
         "comma-spacing": [ 2, { "before": false, "after": true } ],
         "comma-style": [ 2, "last" ],
-        "complexity": [ 2, 4 ],
+        "complexity": [ 2, 8 ],
```
lo1tuma: Please don’t change this rule globally. It is ok to make exception from this rule, but then please only for those functions where it is really necessary via `eslint-disable`-comment.

docs/rules/README.md
```diff
@@ -11,6 +11,7 @@
 * [no-nested-tests](no-nested-tests.md) - disallow tests to be nested within other tests
 * [no-pending-tests](no-pending-tests.md) - disallow pending/unimplemented mocha tests
 * [no-return-and-callback](no-return-and-callback.md) - disallow returning in a test or hook function that uses a callback
+* [no-setup-in-describe](no-setup-in-describe.md) - disallow calling functions and dot operaters directly in describe blocks
```
lo1tuma: Dot operators? Do you mean member expressions? If so, please also consider that member expression could be written as `foo[bar]`.

docs/rules/no-setup-in-describe.md
```diff
@@ -0,0 +1,71 @@
...
+describe('something', function () {
+    const a = setup();
+    const b = a.getter;
+    it('should work', function () {
+        assert(a === b);
+    });
+});
+```
+
+Any setup directly in a `describe` is run before all tests execute. This is undesirable primarily for two reasons:
```
lo1tuma: What about top-level code outside of a `describe`?

docs/rules/no-setup-in-describe.md
```diff
@@ -0,0 +1,71 @@
...
+
+1. When doing TDD in a large codebase, all setup is run for tests that don't have `only` set. This can add a substantial amount of time per iteration.
+2. If global state is altered by the setup of another describe block, your test may be affected.
+
+This rule reports all function calls and use of the dot operator (due to getters and setters) directly in describe blocks.
+
+If you're using [dynamically generated tests](https://mochajs.org/#dynamically-generating-tests), you should disable this rule.
```
lo1tuma: Ah, I was just going to ask about this 😀 .
I think it is ok that this rule will only work for non dynamically generated tests.

lib/rules/no-setup-in-describe.js
```diff
@@ -0,0 +1,100 @@
...
+    additionalSuiteNames = require('../util/settings').additionalSuiteNames;
+
+module.exports = function noSetupInDescribe(context) {
+    var nesting = [],
+        settings = context.settings,
+        FUNCTION = 1,
+        DESCRIBE = 2,
+        PURE = 3;
+
+    function isPureN(node) {
```
lo1tuma: What does `N` mean?

test/rules/pure-describe.js
```diff
@@ -0,0 +1,89 @@
+'use strict';
```
lo1tuma: Why does this has a different filename than the actual rule?

test/rules/pure-describe.js
```diff
@@ -0,0 +1,89 @@
...
+               }
+            },
+            errors: [ {
+                message: 'Unexpected function call in describe block.',
+                line: 1,
+                column: 23
+            } ]
+        },
+        {
+            code: 'describe("", function () { a.b; });',
```
lo1tuma: Can you also add a test for `a['b']`?

.eslintrc
donabrams: This was erroring in plugins other than this one, so I set it to the highest value other plugins required. I can bring this back down and add `eslint-disable-line` to the other plugins if wanted

docs/rules/README.md
donabrams: Ooh, yes, on it!
 - [ ] handle bracket notation (`foo[bar]`)

docs/rules/no-setup-in-describe.md
donabrams: I'd love to handle this, but it could result in worse performance on non-test code (currently a `.length check` is all the overhead that is incurred outside a `describe`). Thoughts on that tradeoff?

donabrams: Hmm, actually I just don't know how to flag a file as having a `describe` with eslint. What's the perf on a `hasDescribe()` and can this even be done?

lib/rules/no-setup-in-describe.js
donabrams: `Node`. I'll expand it.

test/rules/pure-describe.js
donabrams: Oops, changed the name
- [ ] rename this file

test/rules/pure-describe.js
donabrams: - [ ] add test for `foo[bar]`

donabrams:
I've addressed everything I don't have questions about.

.eslintrc
lo1tuma: I’ve fixed that problem in #149. So you just need to rebase from `master`.

docs/rules/no-setup-in-describe.md
lo1tuma: I’m not sure why it should be necessary to check if there is a `describe` in a specific file, you could just look for setup code directly in the `Program` node.
Anyway I’m not sure if this is wanted, because most people probably prefer to have their `require` calls at the top of the file. So maybe we leave it for now as it is and if some one requests that feature we can implement this behind an option.

lo1tuma:
Hi @donabrams, sorry for the late feedback. This looks almost good to me, only a rebase is missing and the removal of the `.eslintrc` changes 😉.

donabrams:
Alright, rebased and complexity "decreased"

PR #174, opened 2018-08-03
Prohibit tests in beforeEach etc. hook calls

papandreou:
This is just a quick implementation to see if there's any interest. And yes, this actually happened :)

Questions:
- [ ] Should this be a separate rule?
- [ ] Is it a requirement to make sure that the error messages say eg. `Unexpected test nested within a beforeEach hook.` rather than `Unexpected test nested within another test.`?

### DIFF
--- a/lib/rules/no-nested-tests.js
+++ b/lib/rules/no-nested-tests.js
@@ -17,16 +17,9 @@ module.exports = function noNestedTests(context) {
         });
     }
 
-    function isNestedTest(isTestCase, isDescribe) {
-        const isNested = testNestingLevel > 0;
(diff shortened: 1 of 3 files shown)


lo1tuma:
Hi @papandreou and thanks for your contribution.
I think this makes definitely sense to be added to this rule.

Regarding the error message it would be of course much better if the message could explain the actual  problem as precisely as possible.

papandreou:
Hey @lo1tuma, thanks for the feedback. I've updated the PR so that separate error messages are issued, when it's a test nested within a hook, or a hook within a hook.

PR #188, opened 2019-02-05
Implement no-async-describe rule

papandreou:
I've seen `describe('my test', async function () { ... })` cause subtle problems, so I think it makes sense to add a rule for it.

### DIFF
--- a/docs/rules/no-async-describe.md
+++ b/docs/rules/no-async-describe.md
@@ -5,7 +5,7 @@ This rule disallows the use of an async function with `describe`. It usually ind
 Example:
 
 ```js
-describe(async function () {
+describe('the thing', async function () {
     // This work should happen in a beforeEach:
     const theThing = await getTheThing();
 
@@ -22,23 +22,26 @@ The rule supports "describe", "context" and "suite" suite function names and dif
 The following patterns are considered problems, whether or not the function uses `await`:
 
 ```js
-describe'something', async function () {
+describe('something', async function () {
     it('should work', function () {});
 });
 
-describe'something', async () => {
+describe('something', async () => {
     it('should work', function () {});
 });
 ```
 
 If the `describe` function does not contain `await`, a fix of removing `async` will be suggested.
 
-This rule looks for every `describe.only`, `it.only`, `suite.only`, `test.only`, `context.only` and `specify.only` occurrences within the source code.
-Of course there are some edge-cases which can’t be detected by this rule, eg.:
+The rule won't be able to detect the (hopefully uncommon) cases where the async
+function is defined before the `describe` call and passed by reference:
 
 ```js
-var describeOnly = describe.only;
-describeOnly.apply(describe);
+async function mySuite() {
+    it('my test', () => {});
+}
+
+describe('my suite', mySuite);
 ```
 
 ## When Not To Use It
--- a/docs/rules/no-async-describe.md
+++ b/docs/rules/no-async-describe.md
@@ -1,0 +1,47 @@
+# Disallow async functions passed to describe (no-async-describe)
+
+This rule disallows the use of an async function with `describe`. It usually indicates a copy/paste error or that you're trying to use `describe` for setup code, which should happen in `before` or `beforeEach`. Also, it can lead to [the contained `it` blocks not being picked up](https://github.com/mochajs/mocha/issues/2975).
+
+Example:
+
+```js
+describe(async function () {
+    // This work should happen in a beforeEach:
+    const theThing = await getTheThing();
+
+    it('should foo', function () {
+        // ...
+    });
+});
+```
+
+## Rule Details
+
+The rule supports "describe", "context" and "suite" suite function names and different valid suite name prefixes like "skip" or "only".
+
+The following patterns are considered problems, whether or not the function uses `await`:
+
+```js
+describe'something', async function () {
+    it('should work', function () {});
+});
+
+describe'something', async () => {
+    it('should work', function () {});
+});
+```
+
+If the `describe` function does not contain `await`, a fix of removing `async` will be suggested.
+
+This rule looks for every `describe.only`, `it.only`, `suite.only`, `test.only`, `context.only` and `specify.only` occurrences within the source code.
+Of course there are some edge-cases which can’t be detected by this rule, eg.:
+
+```js
+var describeOnly = describe.only;
+describeOnly.apply(describe);
+```
+
+## When Not To Use It
+
+- If you use another library which exposes a similar API as mocha (e.g. `describe.only`), you should turn this rule off because it would raise warnings.
+- In environments that have not yet adopted ES6 language features (ES3/5).
--- a/lib/rules/no-async-describe.js
+++ b/lib/rules/no-async-describe.js
@@ -1,0 +1,71 @@
+'use strict';
+
+/* eslint "complexity": [ "error", 5 ] */
+
+/**
+ * @fileoverview Disallow async functions as arguments to describe
+ */
+
+const astUtils = require('../util/ast');
+const { additionalSuiteNames } = require('../util/settings');
+
+module.exports = function (context) {
+    const sourceCode = context.getSourceCode();
+
+    function isFunction(node) {
+        return (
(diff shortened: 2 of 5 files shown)


lib/rules/no-async-describe.js
```diff
@@ -0,0 +1,71 @@
...
+
+    function isFunction(node) {
+        return (
+            node.type === 'FunctionExpression' ||
+            node.type === 'FunctionDeclaration' ||
+            node.type === 'ArrowFunctionExpression'
+        );
+    }
+
+    function containsDirectAwait(node) {
```
papandreou: I don't know if this could be implemented with a funky ramda function that looks for a nested object with a `type` of `AwaitExpression` not contained within one with a `type` of `FunctionExpression` etc. -- but I have to admit that I'm not really familiar with that.

docs/rules/no-async-describe.md
```diff
@@ -0,0 +1,47 @@
+# Disallow async functions passed to describe (no-async-describe)
+
+This rule disallows the use of an async function with `describe`. It usually indicates a copy/paste error or that you're trying to use `describe` for setup code, which should happen in `before` or `beforeEach`. Also, it can lead to [the contained `it` blocks not being picked up](https://github.com/mochajs/mocha/issues/2975).
+
+Example:
+
+```js
+describe(async function () {
```
lo1tuma: Is this a valid example? Doesn’t `describe` require a string as the first argument?

docs/rules/no-async-describe.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+## Rule Details
+
+The rule supports "describe", "context" and "suite" suite function names and different valid suite name prefixes like "skip" or "only".
+
+The following patterns are considered problems, whether or not the function uses `await`:
+
+```js
+describe'something', async function () {
```
lo1tuma: Missing `(` after `describe.

docs/rules/no-async-describe.md
```diff
@@ -0,0 +1,47 @@
...
+    it('should work', function () {});
+});
+
+describe'something', async () => {
```
lo1tuma: Same here.

docs/rules/no-async-describe.md
```diff
@@ -0,0 +1,47 @@
...
+    it('should work', function () {});
+});
+```
+
+If the `describe` function does not contain `await`, a fix of removing `async` will be suggested.
+
+This rule looks for every `describe.only`, `it.only`, `suite.only`, `test.only`, `context.only` and `specify.only` occurrences within the source code.
```
lo1tuma: Why does it look for `it.only` and `test.only` etc? Maybe the whole paragraph can be removed as the first paragraph in this section already explains that `describe`, `context` and `suite` is supported.

docs/rules/no-async-describe.md
```diff
@@ -0,0 +1,47 @@
...
+Of course there are some edge-cases which can’t be detected by this rule, eg.:
+
+```js
+var describeOnly = describe.only;
+describeOnly.apply(describe);
```
lo1tuma: I’m not sure if this example makes sense. It is completely unrelated to `async`.

lib/rules/no-async-describe.js
```diff
@@ -0,0 +1,71 @@
...
+    return {
+        CallExpression(node) {
+            const name = astUtils.getNodeName(node.callee);
+
+            if (astUtils.isDescribe(node, additionalSuiteNames(context.settings))) {
+                const fnArg = node.arguments.slice(-1)[0];
+                if (isAsyncFunction(fnArg)) {
+                    context.report({
+                        node: fnArg,
+                        message: `Do not pass an async function to ${name}()`,
```
lo1tuma: I would suggest something like `Unexpected async function in ${name}()` to be more in line with messages of other rules.

lib/rules/no-async-describe.js
```diff
@@ -0,0 +1,71 @@
...
+            });
+        }
+        return false;
+    }
+
+    function fixAsyncFunction(fixer, fn) {
+        if (!containsDirectAwait(fn.body)) {
+            return fixer.replaceTextRange(
+                [ fn.start, fn.end ],
+                sourceCode.text.slice(fn.range[0], fn.range[1]).replace(/^async /, '')
```
lo1tuma: When a function is async, isn’t the first token always the `async` identifier? If so it might be easier to just get the first token of the function node and use `fixer.remove`.

lib/rules/no-async-describe.js
lo1tuma: Don’t worry, it’s not a requirement to write funky ramda expressions 😉.

test/rules/no-async-describe.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+const RuleTester = require('eslint').RuleTester;
+const rule = require('../../lib/rules/no-async-describe');
+const ruleTester = new RuleTester();
+
+ruleTester.run('no-async-describe', rule, {
+    valid: [
+        'describe()',
+        'describe(function () {})',
```
lo1tuma: The function should be the second argument of describe.

docs/rules/no-async-describe.md
lo1tuma: I guess one edge case that we can’t detect might be something like this:

```js
async function mySuite() {
    it('my test', () => {});
}

describe('my suite', mySuite);
```

papandreou: D'oh, brain fart. I think I copy pasted from https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/test/rules/no-mocha-arrows.js#L15 and then my brain reset itself into a state where `describe` and `it` don't take a string argument.

Fixed in 7c9b6dc71664d8756fca64ddbbf0bf7657f727e7

docs/rules/no-async-describe.md
papandreou: No, it isn't :roll_eyes:

Fixed in 7c9b6dc71664d8756fca64ddbbf0bf7657f727e7

docs/rules/no-async-describe.md
papandreou: Whoops, that part was copy/pasted from the docs for another rule where it did make sense. Maybe I was getting anxious because the other rules are so thoroughly documented :laughing:

lib/rules/no-async-describe.js
papandreou: Fixed in a5fa46a6fdcf798539c11bfedfdce52eeee4aee9. Looks like I copy/pasted from the `no-mocha-global-arrows` rule.

lib/rules/no-async-describe.js
papandreou: That's a good point, fixed in 0c113b53b6a7f82efbd3729b04a7f3c59fbb8f3e

We also need go get rid of the whitespace after `async`, so we can remove the range from the start of the `async` token to the start of the next `function` or `(` token.

test/rules/no-async-describe.js
papandreou: Of course! :+1:

Fixed in 7c9b6dc71664d8756fca64ddbbf0bf7657f727e7

papandreou:
Thanks for the feedback, I think I've addressed all of it now.

By the way, the linting setup for this project was a pleasure to work with. I could really feel the presence of the maintainer's opinions in my editor, and that helped a great deal!

The only *slightly* annoying thing is that I can't even run the test suite if everything doesn't lint, because of https://github.com/lo1tuma/eslint-plugin-mocha/blob/27fa81c76d6274d308c4e6110ff99e73895b7409/package.json#L16 -- that gets in the way when you temporarily have some dead code, or if you've commented out some code without the appropriate amount of whitespace after `//` :)

Could it be changed to `posttest` instead?

lo1tuma:
> The only _slightly_ annoying thing is that I can't even run the test suite if everything doesn't lint

That’s true, I usually use only `npm run test:unit` during development, which should even work in watch mode via `npm run test:unit -- --watch`. This command also doesn’t instrument the code for coverage, so it should be much faster.

PR #190, opened 2019-03-05
Implement no-return-from-async rule

papandreou:
While upgrading a codebase to use async/await I've found it to be a bit confusing when the two Promise styles are mixed so a Promise is returned from an `async` test:

```js
it('should foo', async function () {
  await bar();
  return expect(somePromise).to.eventually.be.rejectedWith('bla');
});
```

I've implemented a rule that prohibits mixing these two styles, similar to to the `no-return-and-callback`. And pretty much all copy/pasted from it because of the similarities :)

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -11,6 +11,7 @@
 * [no-nested-tests](no-nested-tests.md) - disallow tests to be nested within other tests
 * [no-pending-tests](no-pending-tests.md) - disallow pending/unimplemented mocha tests
 * [no-return-and-callback](no-return-and-callback.md) - disallow returning in a test or hook function that uses a callback
+* [no-return-from-async](no-return-from-async.md) - disallow returning from an async test or hook
 * [no-setup-in-describe](no-setup-in-describe.md) - disallow calling functions and dot operators directly in describe blocks
 * [no-sibling-hooks](no-sibling-hooks.md) - disallow duplicate uses of a hook at the same level inside a describe
 * [no-skipped-tests](no-skipped-tests.md) - disallow skipped mocha tests (fixable)
--- a/docs/rules/no-return-from-async.md
+++ b/docs/rules/no-return-from-async.md
@@ -1,0 +1,44 @@
+# Disallow returning from an async test or hook (no-return-from-async)
+
+Mocha's tests or hooks (like `before`) may be asynchronous by returning a Promise. When such a Promise-returning function is defined using [an ES7 `async` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) it can be confusing when combined with an explicit `return` of a Promise, as it's mixing the two styles.
+
+## Rule Details
+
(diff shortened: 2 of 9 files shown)


papandreou:
I guess this could be extended to disallow `return`ing from any `it`, even non-`async` ones. Then you could enforce a code style where all tests that do async work actually are `async` functions.

papandreou:
@lo1tuma, sorry to "at" you, but is this something you would consider adopting?

lo1tuma:
Hi @papandreou, sorry for the late response. I think this rule makes sense. I haven’t had a closer look at the code yet (I hope I find the time soon), but I’m a bit concerned about the copy-pasted code you mentioned in the PR description. Is there any chance to extract and re-use similarities?

papandreou:
@lo1tuma, thanks for taking a look!

I'll see about move the copy/pasted helper functions into `astUtils`.

What do you think about the idea of having a mode where returning from any `it`, even non-async ones is disallowed?

lo1tuma:
> What do you think about the idea of having a mode where returning from any `it`, even non-async ones is disallowed?

What would be the rationale behind it? But now that your are saying it I could imagine a generic `no-return` rule which can be configured to forbid `return` always or just in combination with `callbacks` or `async` functions. That would also solve the duplicated code problem.

papandreou:
@lo1tuma,

> What would be the rationale behind it?

It would be a way to enforce the use of `async`/`await` whenever promises are involved in a test, prohibiting eg.

```
it('should foo', function () {
  return somethingThatReturnsAPromise();
});
```

... while still allowing a value-less (early) return:

```
it('should foo', function () {
  if (bar) {
    return;
  }
  // ...
});
```

... seeing that a promise is the only type of return value from an `it` that `mocha` handles.

So it's really mostly a stylistic thing. I came up with the idea when I did a sweep through a test suite that used promise chains, converting it to `async`/`await` -- I guess the use case for that idea is probably quite narrow :)

papandreou:
I've moved the identical helper functions into `astUtils` now, reducing the duplication quite a lot.

papandreou:
> But now that your are saying it I could imagine a generic `no-return` rule which can be configured to forbid `return` always or just in combination with callbacks or `async` functions. That would also solve the duplicated code problem

My gut feeling is that handling returns in callback-based *and* `async` `it`s would be too much to cram into a single rule. I think the "disallow `return`s of anything but undefined" mode would be more thematically linked to the async case. So I think it should go into `no-return-from-async` (if we want it at all).

papandreou:
> So I think it should go into `no-return-from-async` (if we want it at all).

... Right, except that the naming would be off because the rule would then also affect non-`async` functions :weary:. I think I get why you suggested the `no-return` rule as a compromise :)

lo1tuma:
Maybe it is the best to make small steps. Let’s start with those two separate rules for now and once we have gained some usage experience we can think about if it would make sense to combine them in a generic `no-return` rule.

robertrossmann:
Regarding a rule disallowing `return` completely:

It would be nice, as long as arrow functions with implicit `return` are allowed. Consider:

```js
// Much nice, such arrow
it('works with a promise', () => expect(getpromise()).to.eventually.not.throw(Error))
```

PR #200, opened 2019-07-10
Added recommended ruleset

### LINKED ISSUE #175: Default rules enabled / extends
I have not much knowledge of EsLint, but it is annoying to go through all rules and find out, which to use.

Could not you – just like plain EsLint – somehow also use a thing/things like `"extends": "eslint:recommended"`, so one can get a set of rules that are almost always good for mocha tests?

I think there are many rules that should be enabled "by default"/"by extends".

Lakitna:
Closes #175

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,9 @@ ESLint rules for [mocha](http://mochajs.org/).
 
 This plugin requires ESLint `4.0.0` or later.
 
-`npm install --save-dev eslint-plugin-mocha`
+```bash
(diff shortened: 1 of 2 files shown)


lo1tuma:
Thanks for the PR. The preset looks good to me. It would nice if you could also add a small sentence to the README.

Lakitna:
Good point, I'll add it later.

Lakitna:
Ready for merge on my side :)

PR #209, opened 2019-09-05
Fix no-setup-in-describe to allow Mocha suite config

### LINKED ISSUE #208: no-setup-in-describe: disallows `this.timeout()`/`this.slow()`
As the title says. Is this intentional? I don't see any mention of it in the docs.

If this is considered a bug, I'd be happy to try my hand at a PR to fix!

Thanks for your time :)

straub:
This rule will now allow `this.timeout();`, `this.slow();`, and
`this.retries();` in describe blocks, since they're configuration for
the suite and not setup.

Fixes #208.


test/rules/no-setup-in-describe.js
```diff
@@ -19,10 +19,16 @@ ruleTester.run('no-setup-in-describe', rule, {
         'it("", function () { a[b]; })',
         'it("", function () { a["b"]; })',
         'describe("", function () { it(); })',
+        'describe("", function () { this.slow(1); it(); })',
+        'describe("", function () { this.timeout(1); it(); })',
+        'describe("", function () { this.retries(1); it(); })',
```
lo1tuma: Can you add one test case with a computed member expression? E.g. `this['retries'](1);`

straub: My change to the rule doesn't allow for that form of the expression, so I added this to the invalid test cases, but lmk if you would rather I add support for that form in the rule!

lo1tuma: IMHO we should handle this case as well. There is already a function for that in ast utils: `getPropertyName()`.

straub: All set!

PR #212, opened 2019-09-19
Remove test testCaseNames

perrin4869:
At least in current mocha, `v6`, `global.test` is undefined. And the keyword `test` clashes with functions like:

```javascript
const { test } = require('ramda');
```


lo1tuma:
I can’t find anything in the release notes of mocha v6 regarding this, so maybe this is a bug in mocha? It is also still [documented](https://mochajs.org/#tdd).
Apart from that, `test` still is a global when you explicitly define the TDD interface, e.g. by running `mocha --ui tdd mySpec.js`.

perrin4869:
I see! I will close this one then.
Would you be open to a PR adding an option to exclude certain keywords though?

PR #220, opened 2019-11-22
Support TDD interface in no-setup-in-describe

### LINKED ISSUE #219: [no-setup-in-describe] Does not work with TDD UI
It seems that the rule `no-setup-in-describe` only works with Mocha's `BDD` style and not with `TDD`. For example this snippet does not throw any linting errors

```ts
describe('<Test />', function() {
  afterEach(cleanup);

  it('renders something', function() {
    expect(true).to.equal(true);
  });
});
```

whereas this one

```ts
suite('<Test />', function() {
  teardown(cleanup);

  test('renders something', function() {
    expect(true).to.equal(true);
  });
});
```

throws an `Unexpected function call in describe block  mocha/no-setup-in-describe`. It does not complain when I move `teardown()` up and outside of `suite()` but nesting `teardown()`, `setup()` and so on within `suite()` should [work as well](https://mochajs.org/#tdd)

lo1tuma:
Fixes #219

### DIFF
--- a/lib/util/ast.js
+++ b/lib/util/ast.js
@@ -15,7 +15,10 @@ const describeAliases = [
     'context', 'xcontext', 'context.only', 'context.skip',
     'suite', 'xsuite', 'suite.only', 'suite.skip'
 ];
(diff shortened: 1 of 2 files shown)


screendriver:
LGTM 👍

PR #222, opened 2019-11-22
✉️ Send a webhook to Coveralls when the build completes

screendriver:
This ensures that Coveralls will be notified about the build **after** the last matrix job is finished


lo1tuma:
LGTM 👍

PR #226, opened 2020-01-31
add plugin to recommended config

cruzdanilo:
add plugin to recommended config, so it's not necessary to explicitly add the plugin when using this config.
this is the default for every major eslint plugin with recommended configs, ex:
- `eslint-plugin-node`
- `eslint-plugin-react`
- `eslint-plugin-import`
- `eslint-plugin-jsx-a11y`
- `eslint-plugin-jest`
- `eslint-plugin-promise`
- `eslint-plugin-vue`
- `eslint-plugin-prettier`
- and others.

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,6 +26,7 @@ module.exports = {
     },
     configs: {
         recommended: {
+            plugins: [ 'mocha' ],
             rules: {
                 'mocha/handle-done-callback': 'error',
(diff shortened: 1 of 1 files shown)


brettz9:
In case it provides some assurances, eslint itself also documents this being done for plugins (see the example code which includes `plugins` on each config): https://eslint.org/docs/developer-guide/working-with-plugins#configs-in-plugins

PR #234, opened 2020-02-17
Add schemas for options (and remove for files which are using settings)

brettz9:
If you want me to keep the schemas for settings, I can instead comment them out, but AFAICT, they are only usable with options, not `settings`.

Note that I've had to comment out one test case in each of the two `valid-*-description` rules for this PR as the new schema causes some RegExp literals to fail on a couple properties (since *JSON* Schema can only specify JSON types and does not support indicating a `RegExp` except as a string).

### DIFF
--- a/lib/rules/max-top-level-suites.js
+++ b/lib/rules/max-top-level-suites.js
@@ -11,43 +11,58 @@ const { additionalSuiteNames } = require('../util/settings');
 
 const defaultSuiteLimit = 1;
 
-module.exports = function (context) {
-    const stack = [];
-    const topLevelDescribes = [];
-    const options = context.options[0] || {};
-    const settings = context.settings;
-    let suiteLimit;
-
-    if (isNil(options.limit)) {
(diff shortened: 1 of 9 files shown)


lo1tuma:
> If you want me to keep the schemas for settings, I can instead comment them out, but AFAICT, they are only usable with options, not `settings`.

👍 good catch! Seems like I did a bad job during the code review, since the actual feature was first implemented as rule options and then refactored, due to my review, to settings. Thanks for removing that.

> Note that I've had to comment out one test case in each of the two `valid-*-description` rules

Actually I haven’t noticed that until now. I don’t think this was intentional. It seems like this slipped in while a different feature was added in #206 and it is also not documented. So I would be ok with removing those test-cases and also the whole functionality.

brettz9:
I've amended to remove the regex literal test cases (there were two others but the schema wasn't complaining about them) as well as the detection of RegExp instances. And rebased.

PR #235, opened 2020-02-17
Nondeprecated rule format

brettz9:
Builds on #228 and #234.

Completes move to [non-deprecated](https://eslint.org/blog/2016/07/eslint-new-rule-format) eslint rule format (mentioned first in #228).


brettz9:
This has now been rebased as well.

PR #245, opened 2020-04-23
Add `meta.type` and missing `fixable: 'code'`

brettz9:
add `meta.type` ("problem" or "suggestion") for use by `eslint --fix-type` or formatters and add missing "fixable: code" to `no-skipped-tests`


brettz9:
Note that the failing threshold for the Node 10 test is just due to a minor performance difference (I can bump the threshold if you like).

PR #246, opened 2020-05-13
Updates

brettz9:
Builds on #245 .

- Linting: Check hidden files
- npm: Update devDeps (coveralls, eslint-plugin-node, mocha, nyc)
- Maintenance: Add `.ncurc.js` to prevent peerDep. updates on running local `npm-check-updates`

### DIFF
--- a/.ncurc.js
+++ b/.ncurc.js
@@ -1,0 +1,8 @@
+'use strict';
+
+module.exports = {
+    // Whitelist all for checking besides `peer` which indicates
+    //   somewhat older versions of `eslint` we still support even
+    //   while our devDeps point to a more recent version
+    dep: 'prod,dev,optional,bundle'
+};
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 .idea/**
 build/**
 node_modules/**
+!*.js
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "eslint-plugin-mocha",
-    "version": "6.2.2",
+    "version": "6.3.0",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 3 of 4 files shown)


brettz9:
Again, having a minor issue with the performance budget, so failing one build.

.ncurc.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+module.exports = {
+    // Whitelist all for checking besides `peer` which indicates
```
lo1tuma, line 4: I didn’t know about `npm-check-updates`. I always used `npx david`, but this tool looks more straight forward. Thanks 👍.

brettz9:
Cool, thanks! I've rebased!

brettz9:
Oh--would you like an update for eslint 7 too?

brettz9:
To make it easier for Github Actions, and as many other projects are dropping such support, you might want to drop end-of-lifed Node 8 support and just go to Node 10 (and add 14). I can add that to this PR if you like...

lo1tuma:
Sounds good to me. However if we drop node 8 we should also require eslint > 7 as a peer dependency right?
I would be happy to accept such a PR.

brettz9:
No need to require it as a peer dependency as eslint < 7 still works on Node 10-14. But if you didn't want to add older eslint versions to the Github Actions matrix to ensure multiple eslint versions are still working, and you wanted to ensure no consumers could accidentally use the plugin without warning if using eslint < 7, then we could bump the peerDeps to eslint 7--however you like.

lo1tuma:
You are absolutely right.
But if we are already going to make a breaking change (dropping node 8) we can also combine that with dropping explicit support for older ESLint version (even if we are still compatible with older versions). WDYT?

brettz9:
Sure, SGTM. And as it is only a warning, people could technically use an earlier ESLint version if they wished. I'll add it now on top of the other fixes for ESLint just added.

brettz9:
Added to #247.

PR #253, opened 2020-05-27
Support comments in arrow functions when fixing

### LINKED ISSUE #251: no-mocha-arrows drops comments in the function head when fixing
All of these comments are dropped when fixing:
```js
it( ( foo /* one */ ) /* two */ => /* three */ { body(); } );
```

This was noticed while porting this rule to eslint-plugin-qunit: https://github.com/platinumazure/eslint-plugin-qunit/pull/88

The fixes there can be easily backported to this plugin.

edg2s:
Fixes #251

### DIFF
--- a/lib/rules/no-mocha-arrows.js
+++ b/lib/rules/no-mocha-arrows.js
@@ -15,17 +15,18 @@ module.exports = {
     create(context) {
         const sourceCode = context.getSourceCode();
 
+        function extractSourceTextByRange(start, end) {
+            return sourceCode.text.slice(start, end).trim();
+        }
+
         // eslint-disable-next-line max-statements
         function formatFunctionHead(fn) {
             const arrow = sourceCode.getTokenBefore(fn.body);
-            let firstToken = sourceCode.getFirstToken(fn);
             const beforeArrowToken = sourceCode.getTokenBefore(arrow);
-            let paramsFullText;
-            let params = sourceCode.text.slice(firstToken.range[0], beforeArrowToken.range[1]).trim();
-            let functionKeyword = 'function';
-            const beforeArrowComment = sourceCode.text.slice(beforeArrowToken.range[1], arrow.range[0]).trim();
-            const afterArrowComment = sourceCode.text.slice(arrow.range[1], fn.body.range[0]).trim();
+            let firstToken = sourceCode.getFirstToken(fn);
 
+            let functionKeyword = 'function';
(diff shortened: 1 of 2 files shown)


lib/rules/no-mocha-arrows.js
```diff
@@ -16,21 +15,31 @@ module.exports = {
     create(context) {
         const sourceCode = context.getSourceCode();
 
+        // eslint-disable-next-line max-statements
```
edg2s, line 18: Not sure how you feel about this rule, but I didn't see an obvious way to break up this method without passing multiple return values.

lib/rules/no-mocha-arrows.js
```diff
@@ -16,21 +15,31 @@ module.exports = {
...
-            const paramsRightParen = sourceCode.getTokenBefore(sourceCode.getTokenBefore(fn.body));
-            let paramsFullText = sourceCode.text.slice(paramsLeftParen.range[0], paramsRightParen.range[1]);
+            const arrow = sourceCode.getTokenBefore(fn.body);
+            let firstToken = sourceCode.getFirstToken(fn);
+            const beforeArrowToken = sourceCode.getTokenBefore(arrow);
+            let paramsFullText;
+            let params = sourceCode.text.slice(firstToken.range[0], beforeArrowToken.range[1]).trim();
             let functionKeyword = 'function';
+            const beforeArrowComment = sourceCode.text.slice(beforeArrowToken.range[1], arrow.range[0]).trim();
+            const afterArrowComment = sourceCode.text.slice(arrow.range[1], fn.body.range[0]).trim();
```
lo1tuma, line 27: For better readability I would suggest to extract the repeated pattern of `sourceCode.text.slice(start, end).trim()` to a function.

edg2s: Is
```js
sliceAndTrim( sourceCode.text, beforeArrowToken.range[1], arrow.range[0] );
```
much more concise or legible?

lo1tuma: I think `sourceCode.text` doesn’t need to be a parameter when we define the new function in the scope of the `create` function. Maybe a better name would be something like `extractSourceTextByRange()`. WDYT?

lib/rules/no-mocha-arrows.js
lo1tuma: It’s not a hard requirement. I don’t have a better approach in mind for now. As long as we have a good test coverage we can always refactor the code later 😉.

PR #273, opened 2020-09-27
set env.mocha on recommended config

cruzdanilo:
set `env.mocha: true` on recommended config, so it's not necessary to explicitly set it when using this config.

eslint documents this being done for plugins (the example code includes env settings on each config): https://eslint.org/docs/developer-guide/working-with-plugins#configs-in-plugins

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,6 +27,7 @@ module.exports = {
     },
     configs: {
         recommended: {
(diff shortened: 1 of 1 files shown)


lo1tuma (approved):
LGTM. Thanks.

PR #276, opened 2021-03-05
Speed up no-exclusive-tests and no-pending-tests

szuend:
The `isDescribe` and `isTestCase` are doing a lot of work that is unnecessarily redone on each call expression.
This PR moves this work upfront, when the rule is created and re-uses the result for each call expression.

Please note that the choosen implementation is not a strong opinion. I tried to keep the code churn to a minimum and not export more astUtils internals. Feel free to suggest how you prefer this to be solved differently.

We noticed the performance while linting the [DevTools frontend](https://chromium.googlesource.com/devtools/devtools-frontend/). Running with the following command line:

```sh
TIMING=1 third_party/node/node.py --output scripts/test/run_lint_check_js.js
```

Before the change:
```
Rule                                 | Time (ms) | Relative
:------------------------------------|----------:|--------:
mocha/no-exclusive-tests             |  3697.734 |    35.9%
mocha/no-pending-tests               |  2133.972 |    20.7%
@typescript-eslint/naming-convention |   761.571 |     7.4%
no-whitespace-before-property        |   373.331 |     3.6%
keyword-spacing                      |   359.593 |     3.5%
space-infix-ops                      |   266.812 |     2.6%
@typescript-eslint/func-call-spacing |   180.198 |     1.8%
space-in-parens                      |   175.388 |     1.7%
@typescript-eslint/comma-dangle      |   164.089 |     1.6%
no-trailing-spaces                   |   144.546 |     1.4%
```

After the change (the two eslint-plugin-mocha rules we use no longer show up in the TOP 10):
```
Rule                                 | Time (ms) | Relative
:------------------------------------|----------:|--------:
@typescript-eslint/naming-convention |   758.180 |    17.2%
no-whitespace-before-property        |   348.997 |     7.9%
keyword-spacing                      |   336.037 |     7.6%
space-infix-ops                      |   250.037 |     5.7%
space-in-parens                      |   215.183 |     4.9%
@typescript-eslint/func-call-spacing |   147.527 |     3.4%
@typescript-eslint/comma-dangle      |   144.443 |     3.3%
rulesdir/es_modules_import           |   127.713 |     2.9%
no-trailing-spaces                   |   122.335 |     2.8%
comma-style                          |   112.220 |     2.5%
```

### DIFF
--- a/benchmarks/runtime.bench.js
+++ b/benchmarks/runtime.bench.js
@@ -85,7 +85,7 @@ function lintManyFilesWithAllRecommendedRules({ numberOfFiles }) {
 
 describe('runtime', () => {
     it('should not take longer as the defined budget to lint many files with the recommended config', () => {
-        const budget = 60000000 / cpuSpeed;
+        const budget = 80000000 / cpuSpeed;
 
         const { medianDuration } = runBenchmark(() => {
             lintManyFilesWithAllRecommendedRules({ numberOfFiles: 350 });
--- a/benchmarks/runtime.bench.js
+++ b/benchmarks/runtime.bench.js
@@ -86,7 +86,7 @@ function lintManyFilesWithAllRecommendedRules({ numberOfFiles }) {
 describe('runtime', () => {
     it('should not take longer as the defined budget to lint many files with the recommended config', () => {
         const nodeVersionMultiplier = getNodeVersionMultiplier();
-        const budget = 80000000 / cpuSpeed * nodeVersionMultiplier;
+        const budget = 70000000 / cpuSpeed * nodeVersionMultiplier;
 
         const { medianDuration } = runBenchmark(() => {
             lintManyFilesWithAllRecommendedRules({ numberOfFiles: 350 });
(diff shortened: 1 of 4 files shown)


lo1tuma:
Nice! Thank you for that contribution. I’ve also already started work on a major refactoring of astUtils to also address this and a few other issues, but it is a big change and still not finished yet. So I really appreciate such a small change that improves the performance significantly.

In order to further decrease the performance  I’ve introduce a benchmark test and set a performance budget. Can you check out this benchmark whether we can reduce the budged with this PR? See `https://github.com/lo1tuma/eslint-plugin-mocha/blob/faa8bbf7dcd2a8f22bfe76a3a47b03bb678c2284/benchmarks/runtime.bench.js#L88`.

szuend:
Thanks for the really fast answer! The speed numbers that `os.cpus()` reports for my workstation are fluctuating quite heavily so I choose the new boundary rather conservatively.

I'd like to extend this approach for `isDescribe` and `isTestCase` to other rules that could benefit from them (but not other astUtils functions). Would you prefer a separate PR to keep commits small or should I modify this PR?

lo1tuma (approved):
Thanks. I would say let’s ship those changes. For further performance improvements we can create follow-up PRs.

PR #277, opened 2021-03-05
Improve benchmark test reliability

lo1tuma:
Since older node versions are not as fast as the latest node version we should take that into account and increase the performance budget for older node versions.

cc: @szuend


screendriver (approved):
LGTM 👍

Issue #298, opened 2021-10-10
Support ESLint 8.x


MichaelDeBoey:
ESLint v8.0.0 is [released](https://eslint.org/blog/2021/10/eslint-v8.0.0-released) 🎉

It would be awesome to have official ESLint 8 support. 👊
I'm happy to help where I can of course 🙂

lo1tuma:
Has anyone actually tried the current version `9.0.0` of `eslint-plugin-mocha` with `eslint` v8 and is experiencing any issues?

I don’t think a release is needed for a devDepenency update, as it should not affect the installable package artifact.

lo1tuma:
There is a new version `10.0.0` with updated devDependencies, although I’m pretty sure it doesn’t matter what ESLint version we are using in devDependencies.

PR #299, opened 2021-10-15
Drop support for node v12

lo1tuma:


niksy:
@lo1tuma should this be related to https://github.com/lo1tuma/eslint-plugin-mocha/issues/298#issuecomment-947417958? If Node 12 support is dropped, maybe we should release new major version?

lo1tuma:
@niksy yes this change will be released in the next major version. I don’t think it is related to #298.

WikiRik:
@lo1tuma can I ask why Node 12 completely was dropped? It's still in [maintenance LTS until 2022-04-30](https://nodejs.org/en/about/releases/) and eslint 8 supports `^12.22.0 || ^14.17.0 || >=16.0.0`. Wouldn't it be better to link the support to that?

lo1tuma:
@WikiRik usually I consider to drop support for older node versions every time I’m planning to make a new major-version release. The EoL for node 12 is in only 4 months, so I think it is reasonable for people who still want to use node 12 until then to also use an older version of `eslint-plugin-mocha`.

WikiRik:
@lo1tuma I can see that. Thanks for the quick response!

PR #304, opened 2021-11-13
Add `no-empty-description` rule. closes #302

### LINKED ISSUE #302: Rule proposal: no-empty-description
Hi, that's would be awesome to catch empty suites/tests. 

```js
test('', function () {
    assert.equal(2 + 2, 4);
});
```

Of course, I can use `valid-suite-description`/ `valid-test-description` rules to enforce descriptions format, but I don't have a specific convention, and a lightweight `no-empty` rule will be useful for such a case.

I'm ready to provide help with implementation, if needed.

pustovitDmytro:
This rule enforces specifying the suite/test descriptions for each test. 

The following patterns are considered problems:

```js
it();
suite("");
test(function() { })
test.only(" ", function() { })
```

also, I added this rule to the recommendations list.

### DIFF
--- a/test/rules/no-empty-description.js
+++ b/test/rules/no-empty-description.js
@@ -29,10 +29,6 @@ ruleTester.run('no-empty-description', rules['no-empty-description'], {
         'test.only("some text")',
         'test("some text", function() { })',
 
-        'suite("some text")',
-        'suite.only("some text")',
-        'suite("some text", function() { })',
-
         'notTest()',
 
         {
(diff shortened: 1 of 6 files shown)


lo1tuma (approved):
LGTM. Thank you.

test/rules/no-empty-description.js
```diff
@@ -0,0 +1,87 @@
...
+        {
+            options: [ { testNames: [ 'someFunction' ] } ],
+            code: 'someFunction("this is a test", function () { });'
+        }
+    ],
+
+    invalid: [
+        {
+            code: 'test()',
+            errors: [ { message: defaultErrorMessage, ...firstLine } ],
```
lo1tuma, line 59: We have still `es2017` defined in our `.eslintrc`. Can you update this to `es2019` so the spread operator will be supported? I think this should be save since we dropped node v12.

pustovitDmytro: @lo1tuma fixed in new last commit

lo1tuma (approved):
LGTM

PR #307, opened 2021-12-07
Turn off no-hooks-for-single-case in recommended

### LINKED ISSUE #252: Remove no-hooks-for-single-case from recommended
The number of test in a group may fluctuate over the course of development. While there may be only one test this week, there may be three next week, and back to one the week after. As a developer I wouldn't want to have refactor out the setup/teardown into before/after functions each time I go from have one test to many tests or vice-versa.

Structuring your code to be more flexible for future changes in not really an anti-pattern that warrants inclusion in the recommended config.

edg2s:
Fixes #252


lo1tuma (approved):
Thanks

PR #308, opened 2021-12-08
chore: use `npm-run-all` to run lint/test scripts

bmish:
Adds popular tool [npm-run-all](https://www.npmjs.com/package/npm-run-all) for running scripts.

Benefits:

* Single script to run all types of linting in parallel (`run-p` is alias for `npm-run-all --parallel`)
* Uses double quotes instead of single quotes for better cross-platform support
* Uses npm-run-all instead of `&&` for better cross-platform support
   > Cross platform. We sometimes use & to run multiple command in parallel, but cmd.exe (npm run-script uses it by default) does not support the &. Half of Node.js users are using it on Windows, so the use of & might block contributions. npm-run-all --parallel works well on Windows as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,9 +13,10 @@
         "README.md"
     ],
     "scripts": {
-        "lint:docs": "markdownlint '**/*.md'",
-        "pretest": "eslint .",
-        "test": "npm run test:unit:with-coverage && npm run test:bench",
+        "lint": "run-p --continue-on-error lint:*",
+        "lint:docs": "markdownlint \"**/*.md\"",
+        "lint:js": "eslint .",
+        "test": "npm-run-all test:unit:with-coverage test:bench",
         "test:unit": "mocha test --recursive --reporter dot",
         "test:unit:with-coverage": "nyc npm run test:unit",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -13,9 +13,10 @@
         "README.md"
     ],
     "scripts": {
-        "lint:docs": "markdownlint '**/*.md'",
-        "pretest": "eslint .",
-        "test": "npm run test:unit:with-coverage && npm run test:bench",
+        "lint": "run-p --continue-on-error lint:*",
+        "lint:docs": "markdownlint \"**/*.md\"",
+        "lint:js": "eslint .",
+        "test": "npm-run-all test:unit:with-coverage test:bench",
```
lo1tuma, line 19: This this also run the tasks in parallel?

bmish: This one is just running these two commands in serial. We can change to parallel if you prefer.
```suggestion
        "test": "run-p test:unit:with-coverage test:bench",
```

lo1tuma: Yes, this was also my concern. Thanks for clarifying.

lo1tuma (approved):

PR #310, opened 2021-12-08
Add `mocha` prefix to "no-empty-description" rule name in rec config

JonathanWolfe:
Fixes: Definition for rule 'no-empty-description' was not found


lo1tuma (approved):
Good catch. Thank you!

Issue #311, opened 2021-12-08
new rule `no-empty-description` definition not found in v10 ?


mikehardy:
I see you just released v10 and when I attempted to integrate it, it immediately fails my lint run:

```

/Users/mike/work/Invertase/react-native-google-ads/lib/validateAdRequestOptions.js
  1:1  error  Definition for rule 'no-empty-description' was not found  no-empty-description

/Users/mike/work/Invertase/react-native-google-ads/lib/validateAdShowOptions.js
  1:1  error  Definition for rule 'no-empty-description' was not found  no-empty-description

/Users/mike/work/Invertase/react-native-google-ads/lib/version.js
  1:1  error  Definition for rule 'no-empty-description' was not found  no-empty-description
```

Not sure why, as when I examine node_modules/eslint-plugin-mocha/lib I see the rule, and when I check the code I see that it looks sane.

Very confusing, I might just be doing something wrong, but perhaps others see this?

I successfully linted after adding this to my eslint config even though it makes me sad and I hope it's temporary:

```
    'no-empty-description': 'off',

```

mikehardy:
https://github.com/invertase/react-native-google-ads/ this is the repo it showed up in, it's pretty small, easy to download and run lint on, remove the rule disable, try `yarn lint` etc.

lo1tuma:
This should be fixed in v10.0.1 (see #310).

mikehardy:
Just read the PR, no wonder! Just the prefix. Thanks for helping us keep our code fresh and clean :-)

PR #312, opened 2021-12-08
chore: use `requireindex` to export all rules

bmish:
* No more need to maintain list of exported rules
  * Reduces barrier to creating/renaming lint rules
  * Less chance of typos/mistakes
* [requireindex](https://www.npmjs.com/package/requireindex) is commonly used by other linters for the same purpose, including the generator-eslint blueprint for eslint plugins:
  * https://github.com/eslint/generator-eslint/blob/main/plugin/templates/_plugin.js
  * https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/index.js
  * https://github.com/platinumazure/eslint-plugin-qunit/blob/master/index.js

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,31 +1,9 @@
 'use strict';
 
+const requireIndex = require('requireindex');
+
 module.exports = {
-    rules: {
-        'handle-done-callback': require('./lib/rules/handle-done-callback'),
-        'max-top-level-suites': require('./lib/rules/max-top-level-suites'),
-        'no-async-describe': require('./lib/rules/no-async-describe'),
-        'no-exclusive-tests': require('./lib/rules/no-exclusive-tests'),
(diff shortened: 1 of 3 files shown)


bmish:
Good questions.
* `requireindex` doesn't work with ESM, so if this plugin was converted to ESM, we would need to revert back to manually importing/exporting each rule file
* Fair points about static analysis
* `requireindex` ignores files prefixed with `_`, although I would generally recommend avoiding putting non-rule files in the rules folder (`lib/util` s a good place instead)

This PR is just a suggestion so feel free to decline if you don't think it's worth it. Alternatively, I would recommend adding tests to ensure all rules are correctly exported (to prevent typos/missing rules like in https://github.com/lo1tuma/eslint-plugin-mocha/issues/311).

PR #317, opened 2021-12-30
Document recommended error level

GradedJestRisk:
## :unicorn: Problem
`recommended` rules set include `warning` and `error` level rules.
Some team may rely on rules to prevent unwanted behavior, like the one addressed by `mocha/no-exclusive-tests`.
To make it work in CI, and not to rely on developer's local hooks, the default error level should be elevated from `warning` to `error`.  But to get this property, one has to dig into index.js

## :robot: Solution
Add error level in rules documentation

## :rainbow: Remarks
Using emoticons may be more readable than text (maybe :warning: for `warning` and :bomb:  for `error`?)

## :100: Test
Compare with `index.js`

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -1,27 +1,29 @@
 # Rules
 
-|recommended|fixable|rule|description|
+|recommended (level*)|fixable|rule|description|
 |-|-|-|-|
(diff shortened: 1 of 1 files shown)


lo1tuma (approved):
LGTM. Thanks!

Issue #322, opened 2022-04-12
ramda 0.28.0 regression in 10.0.4 (node 17)


higherorderfunctor:
Plugin is failing to load in a monorepo with an esm module package in it's workspace.  Had to rollback to 10.0.3.  This may be on the ramda side looking at https://github.com/ramda/ramda/issues/3241.

```Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Failed to load plugin 'mocha' declared in '.eslintrc.cjs » ../../.eslintrc.js': Package subpath './src/find' is not defined by "exports" in ./node_modules/ramda/package.json```

./packages/<package>/.eslintrc.cjs
```
module.exports = {
  extends: ["../../.eslintrc.js"],
};
```

./eslintrc.js
```
module.exports = {
  env: {
    es2021: true,
  },
  extends: [
    "airbnb-base",
    "eslint:recommended",
    "plugin:prettier/recommended",
    "plugin:import/recommended",
    "plugin:promise/recommended",
    "plugin:mocha/recommended",
  ],
  plugins: ["prettier", "promise", "mocha"],
  rules: {
    camelcase: "warn",
    "mocha/no-mocha-arrows": "off",
    "no-underscore-dangle": "warn",
    "no-shadow": "off",
  },
};
```

simlu:
Running into the exact same issue here updating various packages. Something definitely broke

lo1tuma:
So it seems to be indeed a problem with ramda itself. Unfortunately the update was necessary since the old version contained a vulnerability, see #320.

[This comment](https://github.com/ramda/ramda/issues/3241#issuecomment-1048184852) suggested a workaround to add the `.js` suffix to all ramda imports. Could anyone confirm that this would fix the issue? If so I would happy to apply this workaround.

higherorderfunctor:
They merged a possible fix a couple hours ago to alter the exports: https://github.com/ramda/ramda/pull/3270

Not sure when it will make it into a release though.

danielduarte:
As a workaround until `ramda` devs release the new version with the fix (that is already merged), I've installed the package pointing to a specific commit and worked fine for me (in Node.js v17 and previous).

    npm install --save-dev "git://github.com/ramda/ramda.git#50c6b57110b2f3631ed8633141f12012b7768d85"

It's not the best, but since we're using `ramda` as an indirect dependency, it solves the issue. And of course, it's temporary.

simlu:
[...]

Issue #319, opened 2022-03-22
Upgrade to ramda 0.28.0


opravil-jan:
Hi,
when do you plan to upgrade ramda dependency to current version. In version 0.27.2 is vulnerability describe here https://github.com/ramda/ramda/pull/3177, please? 

Can I help to do this done.

Thanks
John

lo1tuma:
Fixed by #320

Issue #213, opened 2019-10-03
Performance: plugin adds 150-200ms to ESLint startup time because of ramda


victor-homyakov:
ESLint is frequently used in pre-commit hooks. Just adding `eslint-mocha-plugin` makes linting and commiting noticeably slower even when there is nothing to lint.

**Steps to reproduce**

1. Create clean new folder
2. Install `npm i eslint-mocha-plugin`
3. Create script and execute it in Node.js:
  ```js
const label = 'require plugin';
console.time(label);
const plugin = require('eslint-plugin-mocha');
console.timeEnd(label);
  ```

On MacBook Pro with 2,7 GHz Intel Core i5 it gives me 150-200 ms. In other words, just adding this plugin slows startup of ESLint by 150-200ms.

The problem is in `require('ramda')` - loading ramda module requires that much time. It's OK for long-running applications, but for command-line utilities it is inappropriately slow.

Is it possible to speedup startup time of plugin by rewriting the code without ramda?

lo1tuma:
What makes you think this is caused by ramda? On my machine it takes ~170ms for requiring `eslint-plugin-mocha`, but only ~70ms for requiring `ramda`.

lo1tuma:
I’ve made some optimizations in #214. I think it is not related to the CPU speed, but rather to I/O speed. The ramda entrypoint file does ~250 `require` calls.

victor-homyakov:
> What makes you think this is caused by ramda?

1. I profiled requiring `index.js` - most of the time was spent on `require('./lib/rules/no-exclusive-tests')`
2. Most of the time in `no-exclusive-tests.js` was spent on `const astUtils = require('../util/ast');`
3. Profiling ast utils pointed me to line `const R = require('ramda');`

PR #325, opened 2022-05-02
feat: add support for Node 18

jrandolf:


benchmarks/measure.js
```diff
@@ -2,10 +2,7 @@
 
 const os = require('os');
 const { performance: performanceHooks } = require('perf_hooks');
-const times = require('ramda/src/times');
-const median = require('ramda/src/median');
-const map = require('ramda/src/map');
-const prop = require('ramda/src/prop');
+const { times, median, map, prop } = require('ramda');
```
lo1tuma, line 5: We tried to avoid loading the ramda index file because it caused a significant slowdown of ESLint startup time, see also #213.
I think the root-cause, that this plugin doesn’t work with node 17 and 18 is in ramda itself, see also #322. Unfortunately downgrading ramda seems also like a bad idea because of #319.
The fix in ramda has already been merged, but it hasn’t been released yet, see also https://github.com/ramda/ramda/issues/3278.

taymoork2: Hey, so another route to take instead of waiting for `ramda`, is to switch to [`rambda`](https://github.com/selfrefactor/rambda)
The only limitation in switching to `rambda` is that there is no `memoizeWith` function, so you'll need to implement your own, which isn't difficult
I've already done the work to migrate over, which is based off this PR
Let me know if i should add it to this PR or open a separate PR

Here's a screenshot of the passing tests
image

jrandolf: @lo1tuma Could we push this PR? We are right now blocked on this PR for our support. The "significant slowdown" is significantly less important than supporting Node 18 at this moment.

taymoork2: @jrandolf changes vs my refactor to rambda
image

taymoork2: I'm opening a separate PR @jrandolf

lo1tuma: > The "significant slowdown" is significantly less important than supporting Node 18 at this moment.

@jrandolf this might be the case for you, but probably not for everyone.
Anyway, thanks for taking the effort and trying to help to resolve the issue. I hope to find some time this week to take a closer look at all options and make decision.

PR #326, opened 2022-05-12
Support Node 18 + switch from ramda to rambda

taymoork2:
Thanks to @jrandolf for the original PR #325 this is based off

image

### DIFF
--- a/lib/util/ast.js
+++ b/lib/util/ast.js
@@ -1,11 +1,6 @@
 'use strict';
 
-const complement = require('ramda/src/complement');
-const both = require('ramda/src/both');
-const isNil = require('ramda/src/isNil');
-const propEq = require('ramda/src/propEq');
-const pathEq = require('ramda/src/pathEq');
-const find = require('ramda/src/find');
+const { complement, both, isNil, propEq, pathEq, find } = require('rambda');
 const { getTestCaseNames, getSuiteNames } = require('./names');
 const { getAddtionalNames } = require('./settings');
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                node: [14, 16]
+                node: [14, 16, 18]
         name: Node ${{ matrix.node }}
         steps:
             - uses: actions/checkout@v2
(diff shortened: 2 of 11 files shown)


taymoork2:
@lo1tuma, not be a bother, but can I get a review or update?

taymoork2:
@lo1tuma it's all good

>My first impression is it looks quite good and well maintained. Although it seems like they don’t support importing functions individually. Can you confirm this? Since we don’t use any bundle we won’t profit from thee-shaking.

From what i can see, Rambda, does not support require('rambda/<function>'). The screenshot above show's the runtime with individually imported functions/the PR, I tested using yalp (aka npm link)

lib/util/ast.js
```diff
@@ -50,7 +45,7 @@ function isHookCall(node) {
 function findReference(scope, node) {
     const hasSameRangeAsNode = pathEq([ 'identifier', 'range' ], node.range);
 
-    return find(hasSameRangeAsNode, scope.references);
+    return find(() => hasSameRangeAsNode, scope.references);
```
lo1tuma, line 48: This looks incorrect. The value of `hasSameRangeAsNode` should be a function, which accepts a `reference` object and compares the value of its `identifier.range` path with the value given in `node.value`. But now it is wrapped in another function, ignoring any input value and always returning the function, which is a truethy value.

I wonder why no test is failing.

lo1tuma: I’ve implemented two unit tests here https://github.com/lo1tuma/eslint-plugin-mocha/pull/328 which should fail when `findReference` is not working correctly.

lo1tuma:
@taymoork2 I’ve also checked the result of the [startup benchmark](https://github.com/lo1tuma/eslint-plugin-mocha/blob/6f2c6287cdb841f73bd440aeed2a2ca375185d9c/benchmarks/startup.bench.js#L6) and it seems like `rambda` improves the performance by round about 50%, both regarding runtime duration as well as memory usage.

I’ve only found one small issue, once that is fixed I’m happy to merge this PR.

lib/util/ast.js
taymoork2: Yeah i think i did that because vscode/ts was linting that it was incorrect, but the original implementation was correct, just TS not understanding that pathEq is the correct type
Luckily this doesn't use TSLint so this isn't a problem 🙃

lo1tuma (approved):
LGTM. Thank you!

PR #329, opened 2022-07-14
Support this for additionalCustomNames

thatsmydoing:
A `ThisExpression` doesn't have a name so effectively, `this.it` is parsed as `undefined.it` and that works but is obviously wrong since it would also catch literally `undefined.it` as well.


lo1tuma (approved):
LGTM. Thank you!

PR #336, opened 2023-09-17
Support function calls as custom names (to stop false warnings)

### LINKED ISSUE #335: Support function calls as custom names (to stop false warnings)
### Motivation

Hi. I'd like the following sorts of things to be supported, but they actually give errors in the recommended set:

```javascript
// Say we have a custom function (this one is supposed to ignore tests completely in some situations):
describe.on = (platform) => (process.env.PLATFORM === platform ? describe : (() => {}));

describe('does something', () => {
  before(...);

  // This isn't treated as a suite, so we can get errors about multiple top-level suites, multiple sibling hooks, etc:
  describe.on('windows')('nested', () => {
    before(...); // <- this is treated as a sibling of the above before.
  });
});
```
and
```javascript
const forEach = require('mocha-each');

forEach([ 1, 2, 3 ]).describe('works', (n) => {
  before(...); // <- this is treated as a hook at the top level, even though it's nested here.
  it(...);
});
```

### Possible solution

Allow call expressions to have a `name` when getting the node. It's already the case that dotted names will work and can be specified as custom additional ones.
It's then possible for `additionalCustomNames` to support names of the form
* `{ name: 'describe.on()', type: 'suite' }`,
* `{ name: 'forEach().describe', type: 'suite' }`

so that I can selectively rule these ones *in* to the set of suites.

### Alternatives

* I tried turning off the rule with some `eslint-disable`s but we actually have too many (of the first form I gave in Motivation) to want to do this. And of course, we do want the rules in most places, so globally switching it off isn't suitable.
* In the related PR, my first attempt was to have call expressions with *just* the name, but this didn't work for the `max-top-level-suites` rule (because `describe.on()` and `describe.on()(...)` had the "same" name)

scottmcginness:
Call expressions can have a name when getting the node. Fixes #335 

There might have been better ways to achieve this, but this looked like the most straightforward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,23 +29,46 @@ Then add a reference to this plugin and selected rules in your eslint config:
 
 This plugin supports the following settings, which are used by multiple rules:
 
-* `additionalCustomNames`: This allows rules to check additional function names when looking for suites or test cases. This might be used with a custom Mocha extension, such as [`ember-mocha`](https://github.com/switchfly/ember-mocha)
-**Example:**
+* `additionalCustomNames`: This allows rules to check additional function names when looking for suites or test cases. This might be used with a custom Mocha extension, such as [`ember-mocha`](https://github.com/switchfly/ember-mocha) or [`mocha-each`](https://github.com/ryym/mocha-each).
 
-```json
-{
-    "rules": {
-        "mocha/no-skipped-tests": "error",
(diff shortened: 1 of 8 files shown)


lo1tuma (approved):
LGTM. Thank you!

Issue #185, opened 2018-12-18
Rule proposal: Enforce blank lines around tests


jdoubleu:
It would be nice to have a rule which enforces blank links around `it()` test statements.

I expect the rule to disallow:
```
...
it('should do something', () => {
  ...
})
it('should test other thing', () => {
})
```

But allows:
```
...
it('should do something', () => {
  ...
})

it('should test other thing', () => {
})
```

This is similar to the [`padding-line-between-statements` ESLint rule](https://eslint.org/docs/rules/padding-line-between-statements) but is more specific about the statement (i.e. `it()` function call).

lo1tuma:
Duplicate of #145.

PR #340, opened 2024-01-05
Add `consistent-spacing-between-blocks` rule

bjornua:
This PR introduces a new rule `mocha/consistent-spacing-between-blocks`, which enforces consistent line breaks between Mocha testing blocks (before, after, describe, it, beforeEach, afterEach) within describe blocks.

The following patterns are considered errors:

```javascript
describe("MyComponent", function () {
    beforeEach(function () {
        // setup code
    });
    it("should behave correctly", function () {
        // test code
    });
});
```

These patterns would not be considered errors:

```javascript
describe("MyComponent", function () {
    beforeEach(function () {
        // setup code
    });

    it("should behave correctly", function () {
        // test code
    });

    afterEach(function () {
        // teardown code
    });
});
```

The motivation behind this rule is to improve the readability of test code, making it easier to distinguish different sections. To me, this convention seems very common. It's also used in all the examples in https://mochajs.org/. Happy to hear your thoughts

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,30 +107,31 @@ See [Configuring Eslint](http://eslint.org/docs/user-guide/configuring) on [esli
 ✅ Set in the `recommended` [configuration](https://github.com/lo1tuma/eslint-plugin-mocha#configs).\
 🔧 Automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/user-guide/command-line-interface#--fix).
 
-| Name                                                               | Description                                                             | 💼 | ⚠️ | 🚫 | 🔧 |
-| :----------------------------------------------------------------- | :---------------------------------------------------------------------- | :- | :- | :- | :- |
-| [handle-done-callback](docs/rules/handle-done-callback.md)         | Enforces handling of callbacks for async tests                          | ✅  |    |    |    |
-| [max-top-level-suites](docs/rules/max-top-level-suites.md)         | Enforce the number of top-level suites in a single file                 | ✅  |    |    |    |
-| [no-async-describe](docs/rules/no-async-describe.md)               | Disallow async functions passed to describe                             | ✅  |    |    | 🔧 |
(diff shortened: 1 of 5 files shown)


lo1tuma:
Seems like this is related to #145 and prior work has been done in #185 but it got never finished. Although #145 mentions a few other styling preferences like never having padding between test functions or always having them. Ideally we can support all of them in one rule and having an option that makes the style configurable e.g. `always`, `never` and `between` 🤔.

Apart from that, we also need some unit tests to ensure the rule is working as expected and also to verify edge-cases like how does it behave when there are arbitrary statements between test functions or when test functions are on the same line?

Examples:
```js
// everything in one line, does the rule behave correctly?
describe('', function () { it('', function () {}); it('', function () {}) });

// other code between two test functions, what should the rule do here?
describe('', function () {
  it('', function () {});
  const foo = 'bar';
  it('', function () {});
});
```

bjornua:
> Seems like this is related to #145 and prior work has been done in #185 but it got never finished. Although #145 mentions a few other styling preferences like never having padding between test functions or always having them. Ideally we can support all of them in one rule and having an option that makes the style configurable e.g. `always`, `never` and `between` 🤔.

Adding spacing modes sounds interesting. I'm wondering if we (or someone motivated) could follow up on that in the future? `between` certainly seems like the most popular style and to me a sensible default.

> Apart from that, we also need some unit tests to ensure the rule is working as expected and also to verify edge-cases like how does it behave when there are arbitrary statements between test functions or when test functions are on the same line?

Added test cases. Hopefully the workflows should pass now

lo1tuma:
Sorry for the late response.

> I'm wondering if we (or someone motivated) could follow up on that in the future?

I’ve thought about this and I like this approach. Although introducing more spacing options might require a different rule name which would be a breaking change. That’s being said, I’m willing to make this breaking change at some point. For now, it solves one specific use case and we can learn in future feature requests what exactly is needed.

lo1tuma (approved):
LGTM

PR #343, opened 2024-02-16
Force exact dependency versions

screendriver:


package.json
```diff
@@ -27,24 +27,24 @@
         "release": "release-it"
     },
     "dependencies": {
-        "eslint-utils": "^3.0.0",
-        "rambda": "^7.4.0"
+        "eslint-utils": "3.0.0",
+        "rambda": "7.4.0"
```
lo1tuma, line 31: Please keep the ranges for regular dependencies. So dependency deduplication works for consumers.

screendriver: Done ✅

lo1tuma (approved):

PR #345, opened 2024-02-16
Remove support for node v14, v16 and v18

lo1tuma:


screendriver (approved):

package.json
```diff
@@ -3,7 +3,7 @@
     "version": "10.3.0",
     "description": "Eslint rules for mocha.",
     "engines": {
-        "node": ">=14.0.0"
+        "node": ">=20.0.0"
```
mdmower-csnw, line 6: @lo1tuma - this implies that "eslint-plugin-mocha will only work on Node.js 20+", right? are you really cutting off support for Node 18 which is [supported until April/May 2025](https://nodejs.org/en/about/previous-releases)?

lo1tuma: This change hasn’t been released yet. I’m aware that this is a breaking change and it will land in a new major version.
People who still need node versions < 20 I recommend using `eslint-plugin-mocha` 10.x. If there are any critical bugs I’m happy to backport them to v10 and make new patch releases as long as node v18 is around.

PR #349, opened 2024-03-07
Add support for eslint flat config format

### LINKED ISSUE #341: ESLint 9 / Flat config
Hi,

As of ESLint 8, there is now a new format for ESLint config files which requires that plugins follow a specific format. In ESLint 9, this format will be the only one available by default (and other API changes have occurred in ESLint 9 that may affect this plugin).

Info:

1. https://eslint.org/docs/latest/use/configure/configuration-files-new
1. https://eslint.org/blog/2022/08/new-config-system-part-1/
1. https://eslint.org/blog/2022/08/new-config-system-part-2/
1. https://eslint.org/docs/latest/use/configure/migration-guide

Thanks!

rhysd:
Fix #341

This PR adds support for [the eslint flat config](https://eslint.org/docs/latest/use/configure/configuration-files-new). Now eslint-plugin-mocha can be used in `eslint.config.js`.

```js
import eslint from '@eslint/js';
import mocha from 'eslint-plugin-mocha';

export default [
    eslint.configs.recommended,
    mocha.configs.flat.recommended
];
```

Changes are as follows:

- Since the configuration format is different from legacy eslintrc config, new objects for `all` and `recommended` were added to `mochaPlugin.configs.flat` separately.
- `globals` package was added as dependency because `env` config was dropped at flat config. Globals such as `describe` need to be added in `languageOptions` and `globals` package provides the list of mocha's globals.
- Instruction for flat config was added to README.md document

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ This plugin requires ESLint `4.0.0` or later.
 npm install --save-dev eslint-plugin-mocha
 ```
 
+### `.eslintrc.json`
+
 Then add a reference to this plugin and selected rules in your eslint config:
 
 ```json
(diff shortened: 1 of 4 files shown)


lo1tuma (approved):
LGTM. Thank you!

PR #357, opened 2024-08-19
Use node builtin assert instead of chai

lo1tuma:


benchmarks/runtime.bench.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const { expect } = require('chai');
+const assert = require('assert');
```
screendriver, line 3: Should we use `node:assert` here?

lo1tuma: There are already a bunch of other node builtins being used, without the prefix. So I would make this change in a separate PR with a linting rule enforcing this.

screendriver (approved):

Issue #363, opened 2024-09-25
`no-empty-description` does not detected when in the same line happens also the `no-exclusive-tests`


sokraglav13:
Hello, I found an issue when Im running `npx eslint Test-1.test.js` in the test case.
The issue is that in the same lines breaks two rules but in console prints only the one rule.

### My System Specs
Node: v18.17.0
eslint: v8.57.1
slint-plugin-mocha: v10.5.0

### Reproduce Steps

1) The structure of the test is as below:
`it.only("", async function () {
        ...
    });`
2) Run `npx eslint Test-1.test.js` 
   
### Output

    `error  Unexpected exclusive mocha test  mocha/no-exclusive-test
`
So the question is should find both rules in one line ?
Has anyone identify this behaviour ?

Thank you!

lo1tuma:
Interesting. It seems `no-empty-description` only detects empty descriptions when not using a modifier like `.only()`. So `it("", () => {})` is detected but `it.only("", () => {})` is not detected. That is something we should fix.

Issue #327, opened 2022-05-18
False positive when test is not a mocha global


perrin4869:
Let's say the test file is something like:

```javascript
import { test } from 'ramda';

const validate = test(new RegExp(...), "foo"); // mocha/no-global-tests,mocha/no-pending-tests

describe(..., () => {
  it(..., () => {
    test(new RegExp(...), "foo"); // mocha/no-nested-tests
  });
});
```

The above will be false positives because `test` here is not a mocha global

lo1tuma:
Hello @perrin4869, thanks for raising this issue. This is indeed a problem, in several rules of this plugin. I have already an idea how to fix it and started a while ago with a proof-of-concept implementation, unfortunately I wasn’t able to find enough time to finish it. The change is quite big, and changes basically all rules and how they work, which could also have a lot of impact on the performance of this plugin (I don’t know yet if the impact is negative or positive though). I hope to get the PoC in a shape where I can push it soon.

Related: #296

Issue #264, opened 2020-08-07
isMochaFunctionCall doesn't account for situation where mocha members are being imported from mocha


mwgamble:
This code does not trigger the `mocha/no-exports` rule. It appears this is happening because the `isMochaFunctionCall` function in `lib/util/ast.js` thinks the calls to `describe()` and `it()` are shadowed references, and therefore doesn't mark them as mocha function calls. Any rule relying on this to detect mocha test files (such as `mocha/no-exports`) therefore does not work correctly.

```
import { describe, it } from "mocha";
import { assert } from "chai";

export const rx = 43; 

describe("dummy test suite", function () {
  it("dummy test", function () {
    assert.throws(() => {
      throw new Error("y");
    }); 
  }); 
});

export const x = 42;
```

PR #365, opened 2025-01-22
Implement new resolve algorithm

lo1tuma:
**Breaking Changes:**

- rule `no-skipped-tests` has been removed, its functionality has been merged into the existing `no-pending-tests` rule
- configuring `mocha/additionalCustomNames` now only allows to assign a custom name to exactly one interface
- dropped support for `qunit` interface
- rule `handle-done-callback`: option `ignoreSkipped` has been renamed to `ignorePending`
- rule `valid-suite-description`: the rule option must always be an object
- rule `valid-test-description`: the rule option must always be an object

**Features:**
- new setting `mocha/interface`: when setting this to `exports` (default is `BDD`) then mocha function calls are resolved from import statements (e.g. `import { it } from 'mocha';`)
- variable references are now tracked, e.g. `const foo = it; foo.only('bar', () => {})` is now correctly detected is usage of `it.only()`

Fixes: #363, #327, #264


screendriver (approved):

lib/rules/valid-suite-description.js
```diff
@@ -1,45 +1,20 @@
...
+function objectOptions(options = {}) {
     const {
         pattern: stringPattern,
-        suiteNames = defaultSuiteNames,
         message
     } = options;
 
     const pattern = new RegExp(stringPattern, 'u');
 
-    return { pattern, suiteNames, message };
```
ih1-ac, line 31: The documentation still says `suiteNames` is available, and the changelog for 11.0.0 fails to mention this breaking change.

lo1tuma: It is still possible to define custom suiteNames via the [plugin settings](https://github.com/lo1tuma/eslint-plugin-mocha?tab=readme-ov-file#plugin-settings). It will be respected by all rules, so individual rules don’t have to define this.

SultanWithAGremlin: Still it was not mentioned anywhere. As in my project `context()` is not used in the same way as `describe()` is, it was excluded from the rule by specifying "describe" as the only key for `suiteNames` in the rule. So I now have hundreds of warnings for that.
Simply removing this feature without mentioning it is quite annoying, as the plugin settings just allow for _additionalCustomNames_ but not skipping existing one as possible before.

khitrenovich:
> * rule `no-skipped-tests` has been removed, its functionality has been merged into the existing `no-pending-tests` rule

It would be great to have this documented somewhere. I've opened #403 about it.

Issue #364, opened 2024-11-21
no-setup-in-describe triggers on new before/beforeEach/after/afterEach.timeout


perrin4869:
This was added here: https://github.com/mochajs/mocha/pull/5231
And the typings were just merged: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/71149

```
describe("my suite", () => {
  before(() => {}).timeout(1000); // triggers mocha/no-setup-in-describe
});
```

The above should now be valid (on mocha 11), but is triggering

PR #380, opened 2025-04-30
Add regression tests for no-setup-in-describe

perrin4869:
closes https://github.com/lo1tuma/eslint-plugin-mocha/issues/364
I checked on version 10.5.0, an the code was failing to pass there
Seems version 11 somehow fixed the issue, so just adding a regression test


lo1tuma (approved):
LGTM. Thank you!

